Issue #6, opened 2022-03-26
[Feature] Support notify to telegram bot or shell command


marcuslannister:
If it can not support notify to telegram bot, then support notify to shell command.
Then I can use script like this [[Send message to Telegram on any SSH login - Konstantin Bogomolov | Web Developer](https://bogomolov.tech/Telegram-notification-on-SSH-login/)](https://bogomolov.tech/Telegram-notification-on-SSH-login/) for notification.

Thanks.

haoel:
Look quite straightforward, could you please contribute to this feature?

- Hao

marcuslannister:
I am an embedded system software engineer, I am familiar with C and know a little python, but I am not using go before.

I would like to study go and help to finish this feature, but I can not control the schedule.

haoel:
Okay, let me arrange this requirement in next release. 

BTW, Go is so easy to learn especially for C programmer. 

-- Hao

haoel:
@marcuslannister the telegram notification is supported, close this issue now. if you have further questions, please feel free to reopen this.

avimar:
You might be interested in connecting with caronc/apprise to have an external app handle tons of different integrations without having to write one yourself for each. It's a python code base, though.

PR #7, opened 2022-03-27
feat(notify): support telegram notification and notify code refactory

haoel:
This PR includes the following changes

- Add telegram notify (#6)
- Discord notify can customize the avatar and thumbnail
- Add the `timeout `& `name` for notify.
- Refactory the notify retry code.

### DIFF
--- a/global/global.go
+++ b/global/global.go
@@ -118,18 +118,18 @@ func (t *TLS) Config() (*tls.Config, error) {
 }
 
 // DoRetry is a help function to retry the function if it returns error
-func DoRetry(kind string, tag string, r Retry, fn func() error) error {
+func DoRetry(kind, name, tag string, r Retry, fn func() error) error {
 	for i := 0; i < r.Times; i++ {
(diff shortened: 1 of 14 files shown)


global/global.go
```diff
@@ -110,3 +116,20 @@ func (t *TLS) Config() (*tls.Config, error) {
...
+			return nil
+		}
+		log.Warnf("[%s / %s / %s] Retried to send %d/%d - %v", kind, name, tag, i+1, r.Times, err)
+
+		// last time no need to sleep
+		if i < r.Times-1 {
+			time.Sleep(r.Interval)
+		}
+	}
+	return fmt.Errorf("[%s / %s / %s] failed after %d retries", kind, name, tag, r.Times)
```
zhao-kun, line 134: It would be better to print the detailed information of the last error

haoel: done!

zhao-kun (approved):
LGTM

PR #21, opened 2022-04-02
:sparkles: (Dockfile) Upgrade Dockerfile

xiekeyi98:
1. bump golang from 1.17 to 1.18. 2. use buidkit for docker build.

breaking: build docker with buildkit for performance,but it require
18.09 or
   higher(https://docs.docker.com/develop/develop-images/build_enhancements/)


xiekeyi98:
image

resources/Dockerfile
```diff
@@ -1,7 +1,16 @@
-FROM golang:1.17-alpine as builder
+FROM golang:1.18-alpine as builder
```
haoel, line 1: I remember I updated this yesterday in #18, anyway, not sure why have this diff..

xiekeyi98: sorry, I think cuz I forgot to trace the upstream.

haoel (approved):
thanks for this contribution!

PR #31, opened 2022-04-11
Improve null pointer exception when TLS is not configured

muicoder:
>There are also log level optimizations, and build.


muicoder:
Implemented in probe.

haoel (approved):
LGTM

PR #32, opened 2022-04-11
chore: add arm64 support for docker image

gelleson:
Add support of arm to docker images

### DIFF
--- a/.goreleaser.yaml
+++ b/.goreleaser.yaml
@@ -57,3 +57,16 @@ dockers:
     dockerfile: resources/Dockerfile.goreleaser
     extra_files:
       - resources/scripts/entrypoint.sh
+
+  - image_templates:
+      - megaease/easeprobe:latest
+      - megaease/easeprobe:{{ .Tag }}
+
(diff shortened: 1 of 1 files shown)


haoel:
Thank for this PR @gelleson 

But I don't think it works. I think you need to use the docker manifest. 
You can take a look at this: https://goreleaser.com/customization/docker_manifest/

gelleson:
You are right! Thanks. I will try to rework my PR.

haoel:
Just a couple of hints:

1) For `.goreleaser.yaml` we need to do these changes:
- use docker buildx
- add the `--platform=linux/arm64/v8` as docker build parameters
- add the `docker_manifest` section

2) For  `Dockerfile.goreleaser` we need to deal with the platform. 
- add the `--platform=$BUILDPLATFORM` for `FROM alpine:latest`

gelleson:
Good point. Later I will try to update my PR.

haoel:
Due to this PR is aged, support arm64 docker image in #50

PR #33, opened 2022-04-11
support the bastion/jumper server for SSH probe

haoel:
- Support the bastion/jumper server 
- SSH host support  `user@server:port` scheme
- recursively reflect the probe interface in config structure
- remove the "known_hosts".
- refine the probe log code - remove the duplication code

### DIFF
--- a/probe/ssh/ssh.go
+++ b/probe/ssh/ssh.go
@@ -55,6 +55,19 @@ type SSH struct {
 // BastionMap is a map of bastion
 var BastionMap map[string]Endpoint
 
+// ParseAllBastionHost parse all bastion host
+func ParseAllBastionHost() {
+	for k, v:= range BastionMap {
+		err := v.ParseHost()
+		if err != nil {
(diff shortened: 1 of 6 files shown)


probe/ssh/ssh.go
```diff
@@ -60,12 +67,26 @@ func (s *SSH) Config(gConf global.ProbeSettings) error {
 		return fmt.Errorf("password or private key is required")
 	}
 
+	if len(s.BastionID) > 0 {
+		if bastion, ok := BastionMap[s.BastionID]; ok {
+			log.Debugf("[%s / %s] - has the bastion [%s]", s.ProbeKind, s.ProbeName, bastion.Host)
+			s.bastion = &bastion
+			if err := s.bastion.ParseHost(); err != nil {
```
zhao-kun, line 74: The same bastion endpoint could be parsed repeatedly if it was referenced multiple times. Although it isn't a bug, it's better to fix it.

haoel: this is a good finding,  and fixed it

zhao-kun (approved):
LGTM

PR #35, opened 2022-04-11
feat: add custom timeout for http probe or set default

gelleson:


haoel:
This feature had already been implemented by `probe/base.go` which affects all of the probers. 

there is a `DefaultOptions` which embed in all of the probers, and there is a `ProbeTimeout` & `ProbeInterval` can be configured by all of probers.

PR #39, opened 2022-04-13
feat(notify): support Lark(Feishu) notification.

tigerlee:


haoel:
Thanks for this contribution, I will review the code as soon as possbile.

report/sla.go
```diff
@@ -338,3 +338,66 @@ func SLAStatusText(s probe.Stat, t Format) string {
...
+			},
+			"elements": [%s
+				{
+					"tag": "hr"
+				}, {
+					"tag": "note",
+					"elements": [
+						{
+							"tag": "plain_text",
+							"content": "来自EaseProbe"
```
haoel, line 385:
```suggestion
							"content": global.Prog 
```

report/result.go
```diff
@@ -183,3 +183,62 @@ func ToSlack(r probe.Result) string {
...
+				},
+				{
+					"tag": "hr"
+				},
+				{
+					"tag": "note",
+					"elements": [
+						{
+							"tag": "plain_text",
+							"content": "来自EaseProbe"
```
haoel, line 220:
```suggestion
							"content": global.Prog 
```

notify/lark/lark.go
```diff
@@ -0,0 +1,92 @@
...
+		return err
+	}
+	ret := make(map[string]interface{})
+	err = json.Unmarshal(buf, &ret)
+	if err != nil {
+		return fmt.Errorf("Error response from Lark [%d] - [%s]", resp.StatusCode, string(buf))
+	}
+	if ret["code"] != "0" {
+		return fmt.Errorf("Error response from Lark [%.0f] - [%s]", ret["code"], string(buf))
+	}
```
haoel, line 90: so, whatever success or failure, it always responds to the  200 HTTP Status Code?

haoel (changes requested):

notify/lark/lark.go
tigerlee: hmm, this might not the proper way to handle failure. I'll fix this.

notify/lark/lark.go
```diff
@@ -0,0 +1,94 @@
...
+	ret := make(map[string]interface{})
+	err = json.Unmarshal(buf, &ret)
+	if err != nil {
+		return fmt.Errorf("Error response from Lark [%d] - [%s]", resp.StatusCode, string(buf))
+	}
+	// Server returns {"Extra":null,"StatusCode":0,"StatusMessage":"success"} on success
+	// otherwise it returns {"code":9499,"msg":"Bad Request","data":{}}
+	if ret["StatusCode"] != "0" {
+		return fmt.Errorf("Error response from Lark [%d] - [%s]", int(ret["code"].(float64)), ret["msg"])
+	}
```
haoel, line 92: I mean the HTTP Status Code.  it's  `resp.StatusCode` - the status code for HTTP instead of Lark API's. for examples,  200 means success,  302 means redirect, 403 means permission, 503 means system error.

So, doesn't the Lark use this way to return the success?

tigerlee: When success, server returns HTTP status code `200` with HTTP response body `{"Extra":null,"StatusCode":0,"StatusMessage":"success"}`, so if the `StatusCode` in response body not equals to `"0"`, it's must be a error response, and we can use the `code` field to indicate the error.

So, I use the code in response body to instead of the HTTP status code.

haoel: OK, If you confirm this, I'm going to approve this PR with this uncommon and silly design. :joy:

haoel (approved):
LGTM. Thanks for this contribution!

PR #40, opened 2022-04-13
feat(probe): support host resource usage probe

haoel:
Support the host probe, the configuration example as below.

The feature probe the CPU, Memory, and Disk usage, if one of them exceeds the threshold, then mark the host as status down.

```yaml
host:
  bastion: # bastion server configuration
    aws:
      host: ubuntu@example.com
      key: /path/to/bastion.pem

  servers:
    - name : aws server
      bastion: aws #  <-- bastion server id
      host: ubuntu@172.20.2.202:22
      key: /path/to/server.pem
      threshold:
        cpu: 0.80  # cpu usage  80%
        mem: 0.70  # memory usage 70%
        disk: 0.90  # disk usage 90%

    # using the default threshold - cpu 80%, mem 80%  and disk 95%
    - name : My VPS
      host: user@example.com:22
      key: /Users/user/.ssh/id_rsa
```

### DIFF
--- a/README.md
+++ b/README.md
@@ -372,7 +372,10 @@ Support the host probe, the configuration example as below.
 
 The feature probe the CPU, Memory, and Disk usage, if one of them exceeds the threshold, then mark the host as status down.
 
-> Note: The disk usage only check the root disk.
+> Note: 
+> - The thresholds are **OR** condition, if one of them exceeds the threshold, then mark the host as status down.
+> - The Host needs remote server have the following command: `top`, `df`, `free`, `awk`, `grep`, `tr`, and `hostname` (check the [source code](./blob/work/probe/host/host.go) to see how it works).
+> - The disk usage only check the root disk.
 
 ```yaml
 host:
--- a/README.md
+++ b/README.md
@@ -15,6 +15,7 @@ EaseProbe is a simple, standalone, and lightWeight tool that can do health/statu
     - [3.2 TCP Probe Configuration](#32-tcp-probe-configuration)
     - [3.3 Shell Command Probe Configuration](#33-shell-command-probe-configuration)
     - [3.4 SSH Command Probe Configuration](#34-ssh-command-probe-configuration)
+    - [3.5 Host Resource Usage Probe Configuration](#35-host-resource-usage-probe-configuration)
     - [3.5 Native Client Probe](#35-native-client-probe)
     - [3.6 Notification Configuration](#36-notification-configuration)
     - [3.7 Global Setting Configuration](#37-global-setting-configuration)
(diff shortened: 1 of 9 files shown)


README.md
```diff
@@ -79,7 +80,21 @@ Ease Probe supports the following probing methods:
...
+  host:
+    servers:
+      - name : server
+        host: ubuntu@172.20.2.202:22
+        key: /path/to/server.pem
+        threshold:
+          cpu: 0.80  # cpu usage  80%
+          mem: 0.70  # memory usage 70%
+          disk: 0.90  # disk usage 90%
+  ```
```
zhao-kun, line 95: Need to add a description to explain the relationship between thresholds condition. From my understanding, it's the `AND` relationship. Notification will be sent only when all meters are greater than the threshold.

probe/host/host.go
```diff
@@ -0,0 +1,231 @@
...
+	// 2. retrive the os:  			`awk -F= '/^NAME/{print $2}' /etc/os-release | tr -d '\"'`
+	// 3. retrive the memory usage:	`free -m | awk 'NR==2{printf "%s %s %.2f\n", $3,$2,$3*100/$2 }'`
+	// 4. retrive the cpu core:		`grep -c ^processor /proc/cpuinfo;`
+	// 5. retrive the cpu usage:	`top -b -n 1 | grep Cpu | awk -F ":" '{print $2}'`
+	// 6. retrive the disk usage	`df -h | awk '$NF=="/"{printf "%d %d %s\n", $3,$2,$5}'`
+	s.Command = `hostname;
+	awk -F= '/^NAME/{print $2}' /etc/os-release | tr -d '\"';
+	free -m | awk 'NR==2{printf "%s %s %.2f\n", $3,$2,$3*100/$2 }';
+	grep -c ^processor /proc/cpuinfo;
+	top -b -n 1 | grep Cpu | awk -F ":" '{print $2}';
```
zhao-kun, line 81: `Top` command is required. Why didn't calculate the CPU usage from `/proc/stat` [1] with the pure script?

[1]https://stackoverflow.com/questions/11356330/how-to-get-cpu-usage

README.md
haoel: I think the `OR` conditions would make more sense. I will state that more clearly

probe/host/host.go
haoel: Using `/proc/stat` to calculate the CPU usage is so complicated, and there are many different methods to do that. And I can see different calculations have different results. I think it might be better to align with the actual operation - when people check the CPU usage, everyone uses `top` instead of `/proc/stat`,  I think using `top` can have the same picture with the operation, and avoid the confusion. That's why I decide to use `top` instead of `/proc/stat`. 

[1] https://stackoverflow.com/questions/9229333/how-to-get-overall-cpu-usage-e-g-57-on-linux (there are many calculations in the answers replies)
[2] https://supportcenter.checkpoint.com/supportcenter/portal?eventSubmit_doGoviewsolutiondetails=&solutionid=sk65143
[3] Another good article to compare the `top` and `/proc/stat` https://medium.com/@yogita088/how-to-calculate-cpu-usage-proc-stat-vs-top-e74f99f02d08

zhao-kun: Leveraging the top command brings a consistent result, I totally agree with you, but it needs an extra component, this is a trade-off. So we should clarify that the probe based on infrastructure metrics needs to ensure the machine has installed the top command.

haoel: I will clarify that in the document.

zhao-kun (approved):

PR #42, opened 2022-04-19
Add the probe `timeout` in README.md

haoel:


zhao-kun (commented):
Grammarly suggests that `All probes` is better than `All of prob`. I find some explanations for this usage [1].

[1] https://getproofed.com/writing-tips/word-choice-all-of/

zhao-kun (approved):

PR #53, opened 2022-04-28
[Feature] Support custom prefered response code for HTTP probe

douglarek:


haoel:
@douglarek  Thanks for this contribution. My comments are as below:

It's better to use a range instead of only one status code. because 1xx , 2xx, 3xx, ... all of them represent success. 

I think we can check the ranges something like:

```yaml
success_code:
   - [200, 210]  # from 200 to 210
   - [300, 320] # from 300 to 320 
```
So, the default range is  [0 - 499]

douglarek:
> @douglarek Thanks for this contribution. My comments are as below:
> 
> It's better to use a range instead of only one status code. because 1xx , 2xx, 3xx, ... all of them represent success.
> 
> I think we can check the ranges something like:
> 
> ```yaml
> success_code:
>    - [200, 210]  # from 200 to 210
>    - [300, 320] # from 300 to 320 
> ```
> 
> So, the default range is [0 - 499]

thanks for the deep thought, @haoel , I will make some minor fix.

probe/http/http.go
```diff
@@ -89,13 +90,19 @@ func (h *HTTP) Config(gConf global.ProbeSettings) error {
 		h.Method = "GET"
 	}
 
+	for _, r := range h.SuccessCode {
+		if len(r) != 2 {
+			log.Errorf("HTTP Success Code is invalid - %+v", r)
+			return fmt.Errorf("HTTP Success Code is invalid - %+v", r)
+		}
+	}
+
```
haoel, line 99: if the `SuccessCode` is empty, we can add the default range [0-499] into the SuccessCode. Then you don't need the if-else in `DoProbe()` function.

probe/http/http.go
```diff
@@ -133,6 +140,17 @@ func (h *HTTP) DoProbe() (bool, string) {
...
+		} else if len(h.SuccessCode) > 0 {
+			var hitted bool
+			for _, r := range h.SuccessCode {
+				if r[0] <= resp.StatusCode && resp.StatusCode <= r[1] {
+					hitted = true
+					break
+				}
+			}
+			if !hitted {
+				return false, fmt.Sprintf("HTTP Status Code is %d, you want %v", resp.StatusCode, h.SuccessCode)
```
haoel, line 152:
```suggestion
				return false, fmt.Sprintf("HTTP Status Code is %d. It missed in %v", resp.StatusCode, h.SuccessCode)
```

probe/http/http.go
```diff
@@ -133,6 +140,17 @@ func (h *HTTP) DoProbe() (bool, string) {
 		if resp.StatusCode >= 500 {
 			message = fmt.Sprintf("HTTP Status Code is %d", resp.StatusCode)
 			status = false
+		} else if len(h.SuccessCode) > 0 {
+			var hitted bool
```
haoel, line 144: the `hitted` wrong word, it should be `hit`. But I think the `valid` would make sense.

probe/http/http.go
```diff
@@ -89,13 +90,19 @@ func (h *HTTP) Config(gConf global.ProbeSettings) error {
 		h.Method = "GET"
 	}
 
+	for _, r := range h.SuccessCode {
+		if len(r) != 2 {
+			log.Errorf("HTTP Success Code is invalid - %+v", r)
+			return fmt.Errorf("HTTP Success Code is invalid - %+v", r)
+		}
```
haoel, line 97: can we just remove the invalid configuration and continue the work?

haoel:
by the way, I think we can remove the `>= 500` checking...

probe/http/http.go
```diff
@@ -116,25 +130,29 @@ func (h *HTTP) DoProbe() (bool, string) {
...
 	resp, err := h.client.Do(req)
 
-	status := true
-	message := ""
 	if err != nil {
-		message = fmt.Sprintf("Error: %v", err)
 		log.Errorf("error making get request: %v", err)
-		status = false
+		return false, fmt.Sprintf("Error: %v", err)
 	} else {
```
haoel, line 136: we can remove the `else`, because the `if {}` returns

probe/http/http.go
```diff
@@ -116,25 +130,29 @@ func (h *HTTP) DoProbe() (bool, string) {
...
-		if resp.StatusCode >= 500 {
-			message = fmt.Sprintf("HTTP Status Code is %d", resp.StatusCode)
-			status = false
+		if !valid {
+			return false, fmt.Sprintf("HTTP Status Code is %d. It missed in %v", resp.StatusCode, h.SuccessCode)
 		}
 	}
 
-	return status, message
+	return true, ""
```
haoel, line 157:
```suggestion
	return true, fmt.Sprintf("HTTP Status Code is %d", resp.StatusCode)
```

probe/http/http.go
```diff
@@ -89,13 +91,25 @@ func (h *HTTP) Config(gConf global.ProbeSettings) error {
...
+			if len(r) != 2 {
+				log.Warnf("HTTP Success Code is invalid - %+v, use default [0, 499]", r)
+				h.SuccessCode = [][]int{{0, 499}}
+				break
+			}
+		}
+	} else {
+		log.Debug("HTTP Success Code is not set, use default [0, 499]")
+		h.SuccessCode = [][]int{{0, 499}}
+	}
```
haoel, line 105: I am thinking of something like the below:

```go
var codeRange [][]int 
for _, r := range h.SuccessCode {
    if len(r) >= 2 {
        codeRange = append(codeRange,  []int{r[0], r[1]} )
    }
}
if len(codeRange) == 0 {
     codeRange =  [][]int{{0, 499}}
}
h.SuccessCode = codeRange

```

README.md
```diff
@@ -277,6 +277,10 @@ http:
     ca: /path/to/file.ca
     cert: /path/to/file.crt
     key: /path/to/file.key
+    # Optional HTTP response code range you want, only HTTP standard codes(>=200 and < 500) work, default is [0, 499].
+    success_code:
+      - [200,210]
+      - [300,320]
```
haoel, line 283:
```suggestion
    # HTTP successful response code range, default is [0, 499].
    success_code:
      - [200,206] # the code >=200 and <= 206
      - [300,308] # the code >=300 and <= 308
```

haoel (approved):
Thanks for this code change!

PR #60, opened 2022-04-29
feat: add sms alert support for Twilio, Vonage, and YunPian

nullsimon:


haoel:
Thanks for this contribution. I have a couple of questions:

- Can we use "batch send" for multiple phone numbers?
- The SLA report could be a very long text, do you think that is suitable for SMS?

Additionally, because there are many SMS vendors, so, if we can do this abstractly and `YunPian` as one of the SMS drivers, that would be wonderful.  Can you do this abstraction?

nullsimon:
> Thanks for this contribution. I have a couple of questions:
> 
> * Can we use "batch send" for multiple phone numbers?
> * The SLA report could be a very long text, do you think that is suitable for SMS?
> 
> Additionally, because there are many SMS vendors, so, if we can do this abstractly and `YunPian` as one of the SMS drivers, that would be wonderful. Can you do this abstraction?

Absolutely yes.

1 "batch send" is a better option for team
2 SLA report is not suitable for SMS, is too expensive, a SMS with link to a web page is a better choice. 
3 for the sms driver abstraction, there are many other good services provider, like twilio, I will do it.

haoel:
Just a couple of hints.

1) For the abstraction, you can talk a look at how the `probe/client/client.go` does.  `

2) For the SLA report, I think we can just give a very short summary like the code in `SLASlack()`

```go
	sla := 0.0
	for _, p := range probers {
		sla += SLAPercent(p.Result())
	}
	sla /= float64(len(probers))
	summary := fmt.Sprintf("Total %d Services, Average %.2f%% SLA", len(probers), sla)
```

nullsimon:
> Just a couple of hints.
> 
> 1. For the abstraction, you can talk a look at how the `probe/client/client.go` does.  `
> 2. For the SLA report, I think we can just give a very short summary like the code in `SLASlack()`
> 
> ```go
> 	sla := 0.0
> 	for _, p := range probers {
> 		sla += SLAPercent(p.Result())
> 	}
> 	sla /= float64(len(probers))
> 	summary := fmt.Sprintf("Total %d Services, Average %.2f%% SLA", len(probers), sla)
> ```

I'm working on it

nullsimon:
> Just a couple of hints.
> 
> 1. For the abstraction, you can take a look at how the `probe/client/client.go` does.  `
> 2. For the SLA report, I think we can just give a very short summary like the code in `SLASlack()`
> 
> ```go
> 	sla := 0.0
> 	for _, p := range probers {
> 		sla += SLAPercent(p.Result())
> 	}
> 	sla /= float64(len(probers))
> 	summary := fmt.Sprintf("Total %d Services, Average %.2f%% SLA", len(probers), sla)
> ```

hi, after I dig into the client implements, I have some problems

I found that different sms providers have different credentials definition, they are not the same, so I should put those provider self only keys into their own business config. 

And business config should be self definition, should be easily understand and see from outside. So should I put those different provider just like different notify, So maybe I don't need abstraction. 

But I know those providers have something in common, If I use the abstraction scheme like client, I have to parse the self only config keys in each file, but I think it is a little dirty. Like this one:

```go
func Plugin(config *viper.Viper) (model.Provider, error) {
	return Config{
		Key:           config.GetString("key"),
		Secret:        config.GetString("secret"),
		EnableWebhook: config.GetBool("webhook.enable"),
	}.New(name)
}
```
[参考代码](https://github.com/minchao/smsender/blob/master/smsender/providers/nexmo/provider.go)

notify/sms/yunpian/yunpian.go
```diff
@@ -0,0 +1,64 @@
...
+}
+
+// Kind return the type of Notify
+func (c Yunpian) Kind() string {
+	return Kind
+}
+
+// Notify return the type of Notify
+func (c Yunpian) Notify(title, text string) error {
+	api := "https://sms.yunpian.com/v2/sms/batch_send.json"
```
haoel, line 34: I think you can use `c.ApiUrl` here.

notify/sms/conf/conf.go
```diff
@@ -0,0 +1,116 @@
...
+
+// Options implements the configuration for native client
+type Options struct {
+	base.DefaultNotify `yaml:",inline"`
+
+	ProviderType ProviderType `yaml:"provider"`
+	Mobile       string       `yaml:"mobile"`
+	From         string       `yaml:"from"`
+	ApiKey       string       `yaml:"apikey"`
+	ApiSecret    string       `yaml:"api_secret"`
```
haoel, line 60: looks `ApiSecret` is only used by Twilio， I think you can move this to Twilio struct. ;-)

notify/sms/conf/conf.go
```diff
@@ -0,0 +1,116 @@
...
+	ApiUrl       string       `yaml:"api_url"`
```
haoel, line 61: let's use short names here: `Key`, `Secret`, and `URL` are clear enough.

haoel (commented):
Hi @nullsimon, 

The code structure is quite good enough. And I made some suggestions you can consider. 

Thanks,
-- Hao

notify/sms/conf/conf.go
nullsimon: > looks `ApiSecret` is only used by Twilio， I think you can move this to Twilio struct. ;-)

I have a problem here, because I don't know how to construct `ApiSecret` in Twilio struct.

nullsimon: And `ApiSecret` is using by Nexmo, `AUTH_TOKEN` is using by Twilio. Maybe we can just using key and secret to abstract them.

```sh
curl -X "POST" "https://rest.nexmo.com/sms/json" \
  -d "from=$VONAGE_BRAND_NAME" \
  -d "text=A text message sent using the Vonage SMS API" \
  -d "to=$TO_NUMBER" \
  -d "api_key=$VONAGE_API_KEY" \
  -d "api_secret=$VONAGE_API_SECRET"
```

```sh
curl -X POST https://api.twilio.com/2010-04-01/Accounts/$TWILIO_ACCOUNT_SID/Messages.json \
--data-urlencode "Body=Hi there" \
--data-urlencode "From=+15017122661" \
--data-urlencode "To=+15558675310" \
-u $TWILIO_ACCOUNT_SID:$TWILIO_AUTH_TOKEN
```

haoel: there are two kinds of construction:

```yaml
sms:
-  name "xxxx"
    driver: "twilio"
    ....
- name "yyy"
   driver: "yunpian"
   ...
```

and 

```yaml
sms:
    twilio:
    - name xxx
       ....
     yunpian:
     - name yyyy
        ...
```
- For the first one, you have to have a common struct that includes all of the fields of all of the SMS providers.
- For the second one, you can embed the common struct into all of the providers' structs. 

Both of ways are fine to me, you can make your own decision.

haoel: But if we go the 2nd way, we have to refractory the `AllNotifiers()` function in  `conf/conf.go`, because we need to find the notification structure recursively... this probably is a bit complicated. 

So, may you can stick to your current way, that would be fewer code changes.

README.md
```diff
@@ -131,6 +131,7 @@ Ease Probe supports the following notifications:
 - **DingTalk**. Support the DingTalk notification.
 - **Lark**. Support the Lark(Feishu) notification.
 - **Log File**. Write the notification into a log file
+- **Sms**. Support Multiple Sms Service Provider
```
haoel, line 134:
```suggestion
- **SMS**. Support SMS notification with multiple SMS service providers
```

README.md
```diff
@@ -172,6 +173,12 @@ notify:
   lark:
     - name: "lark alert service"
       webhook: "https://open.feishu.cn/open-apis/bot/v2/hook/d5366199-xxxx-xxxx-bd81-a57d1dd95de4"
+  sms:
+    - name: "sms alert service"
+      provider: "yunpian"
+      key: xxxxxxxxxxxx # yunpian apikey 
+      mobile: 18800000000,18000000000 # mobile phone number, multi phone number joint by ,
+      sign: "【megaease】" # need to register; usually brand name
```
haoel, line 181:
```suggestion
      sign: "xxxxxxxx" # need to register; usually brand name
```

README.md
```diff
@@ -540,6 +547,14 @@ notify:
     - name: "Local Log"
       file: "/tmp/easeprobe.log"
       dry: true
+  # Notify by sms using yunpian  https://www.yunpian.com/official/document/sms/zh_cn/domestic_single_send
+  sms:
+    - name: "sms alert service"
+      provider: "yunpian"
+      key: xxxxxxxxxxxx # yunpian apikey 
+      mobile: 18800000000,18000000000 # mobile phone number, multi phone number joint by ,
+      sign: "【megaease】" # need to register; usually brand name
```
haoel, line 556:
```suggestion
      mobile: 123456789,987654321 # mobile phone number, multi phone number joint by `,`
      sign: "xxxxx" # get this from yunpian
```

README.md
```diff
@@ -172,6 +173,12 @@ notify:
   lark:
     - name: "lark alert service"
       webhook: "https://open.feishu.cn/open-apis/bot/v2/hook/d5366199-xxxx-xxxx-bd81-a57d1dd95de4"
+  sms:
+    - name: "sms alert service"
+      provider: "yunpian"
+      key: xxxxxxxxxxxx # yunpian apikey 
+      mobile: 18800000000,18000000000 # mobile phone number, multi phone number joint by ,
```
haoel, line 180:
```suggestion
      mobile: 123456789,987654321 # mobile phone number, multiple phone number joint by `,`
```

notify/sms/sms.go
```diff
@@ -0,0 +1,73 @@
...
+func (c *NotifyConfig) Config(gConf global.NotifySettings) error {
+	c.DefaultNotify.Config(gConf)
+	c.configSmsDriver()
+	c.SendFunc = c.DoNotify
+
+	log.Debugf("[%s] configuration: %+v, %+v", c.NotifyStat, c, c)
+	return nil
+}
+
+func (c *NotifyConfig) configSmsDriver() {
```
haoel, line 53: let's change the name `configSmsDriver()` to `configSMSDriver()`

notify/sms/twilio/twilio.go
```diff
@@ -0,0 +1,65 @@
...
+		return err
+	}
+	defer resp.Body.Close()
+
+	buf, err := ioutil.ReadAll(resp.Body)
+	if err != nil {
+		return err
+	}
+	if resp.StatusCode != 200 {
+		return fmt.Errorf("Error response from Sms [%d] - [%s]", resp.StatusCode, string(buf))
```
haoel, line 62:
```suggestion
		return fmt.Errorf("Error response from SMS [%d] - [%s]", resp.StatusCode, string(buf))
```

notify/sms/yunpian/yunpian.go
```diff
@@ -0,0 +1,64 @@
...
+		return err
+	}
+	defer resp.Body.Close()
+
+	buf, err := ioutil.ReadAll(resp.Body)
+	if err != nil {
+		return err
+	}
+	if resp.StatusCode != 200 {
+		return fmt.Errorf("Error response from Sms [%d] - [%s]", resp.StatusCode, string(buf))
```
haoel, line 61:
```suggestion
		return fmt.Errorf("Error response from SMS [%d] - [%s]", resp.StatusCode, string(buf))
```

README.md
```diff
@@ -540,6 +547,14 @@ notify:
     - name: "Local Log"
       file: "/tmp/easeprobe.log"
       dry: true
+  # Notify by sms using yunpian  https://www.yunpian.com/official/document/sms/zh_cn/domestic_single_send
+  sms:
+    - name: "sms alert service"
```
haoel, line 552:
```suggestion
    - name: "sms alert service - yunpian"
```

notify/sms/sms.go
```diff
@@ -0,0 +1,73 @@
...
+	Provider conf.Provider `yaml:"-"`
+}
+
+// Kind return the probe kind
+func (c *NotifyConfig) Kind() string {
+	return c.MyKind
+}
+
+// Config Sms Config Object
+func (c *NotifyConfig) Config(gConf global.NotifySettings) error {
```
haoel, line 44: In this function, we need to initialize the `Format` defined in `report/types.go`, because the Format would be bound with the SLA function, and we know the SLA report is too long for  SMS, we need the SLA summary I mentioned before.

So, I suggest creating a new format, likes 'SMS', and you can bind the related function as below.

```
var FormatFuncs = map[Format]FormatFuncStruct{
   ...
   SMS : {ToText, SLASummary},
}
```

haoel (changes requested):

notify/sms/sms.go
```diff
@@ -0,0 +1,73 @@
...
+	c.DefaultNotify.Config(gConf)
+	c.configSmsDriver()
+	c.SendFunc = c.DoNotify
+
+	log.Debugf("[%s] configuration: %+v, %+v", c.NotifyStat, c, c)
```
haoel, line 49: This log.Debugf() seems wrong...

nullsimon: Let me see, and I'm working with the SLASummary, wait a minute

nullsimon: > This log.Debugf() seems wrong...

try again

haoel (commented):
One more thing, may I ask you go to `SLASlack()` function in `report/sla.go` replace the first 6 lines code with your `SLASummary()` ? 

Thanks in advance!

nullsimon:
> One more thing, may I ask you go to `SLASlack()` function in `report/sla.go` replace the first 6 lines code with your `SLASummary()` ?
> 
> Thanks in advance!

absolutely

README.md
```diff
@@ -131,6 +131,7 @@ Ease Probe supports the following notifications:
 - **DingTalk**. Support the DingTalk notification.
 - **Lark**. Support the Lark(Feishu) notification.
 - **Log File**. Write the notification into a log file
+- **SMS**. Support SMS notification with multiple SMS service providers
```
haoel, line 134:
```suggestion
- **SMS**. Support SMS notification with multiple SMS service providers - [Twilio](https://www.twilio.com/sms), [Vonage](https://developer.vonage.com/messaging/sms/overview), [YunPain](https://www.yunpian.com/doc/en/domestic/list.html)
```

haoel (approved):
Thanks a lot for this good contribution!

nullsimon:
> Thanks a lot for this good contribution!

Thanks for your patience and guidance!

PR #61, opened 2022-04-29
Redirect `df` stderr to /dev/null and limit to rootfs `/`

proditis:
Started reading the code to get a better understanding of the application and i noticed that the df command can be improved a bit.
this change makes df stop reporting errors for fuse filesystems such as `df: /run/user/1000/doc: Operation not permitted` and also limits the usage reported by df to only the root filesystem `df -h /`


haoel (approved):
that's cool!

PR #62, opened 2022-04-29
the api url doesnt work with a / at the end

proditis:
the api url doesnt work with a / at the end

### DIFF
--- a/README.md
+++ b/README.md
@@ -11,7 +11,7 @@ EaseProbe is a simple, standalone, and lightWeight tool that can do health/statu
     - [1.1 Probe](#11-probe)
     - [1.2 Notification](#12-notification)
     - [1.3 Report](#13-report)
-  - [2. Getting Start](#2-getting-start)
+  - [2. Getting Started](#2-getting-started)
     - [2.1 Build](#21-build)
     - [2.2 Run](#22-run)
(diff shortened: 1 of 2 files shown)


haoel:
Good catch!

And I think we can correct the code a bit, just add a trailing slash, change `/api/v1/sla` to `/api/v1/sla/`

Then, access `/api/v1/sla` would be 301 redirected to `/api/v1/sla/`

```patch
diff --git web/server.go web/server.go
index ef01cb9..e3d667c 100644
--- web/server.go
+++ web/server.go
@@ -87,7 +87,7 @@ func Server() {
        // Start the http server
        go func() {
                http.HandleFunc("/", slaHTML)
-               http.HandleFunc("/api/v1/sla", slaJSON)
+               http.HandleFunc("/api/v1/sla/", slaJSON)
                log.Infof("[Web] HTTP server is listening on %s:%s", host, port)
                if err := http.ListenAndServe(host+":"+port, nil); err != nil {
                        log.Errorf("[Web] HTTP server error: %s", err)
```

Could you please make this code change in this PR?

proditis:
Awesome, yes that is far better solution. 

update and pushed. 

Thanks for letting me get my hands-on this :smiley:

haoel (approved):

PR #67, opened 2022-05-03
Fix typo

Allenxuxu:
These errors make it not pass unit tests, and I can add automatic tests to Github Action if needed.


haoel:
good catch! 

And yes, you can add the GitHub action in another PR. 

Thanks,
Hao

haoel (approved):

PR #73, opened 2022-05-09
Use discord username

proditis:
Hi, I had 3 different easeprobe instances running on 3 different hosts monitoring the same targets (for triangulation purposes) but when the alerts on discord arrived i couldnt distinguish which is which since all of them had the same name. 

This change adds support for setting the username that the discord notifications will appear from 
image

```yaml
    - name: "easeprobe #Alert"
      username: "my bot username"
      webhook: "https://discord.com/api/webhooks/.../..."
```

### DIFF
--- a/README.md
+++ b/README.md
@@ -181,7 +181,7 @@ notify:
   sms:
     - name: "sms alert service"
       provider: "yunpian"
-      key: xxxxxxxxxxxx # yunpian apikey 
+      key: xxxxxxxxxxxx # yunpian apikey
       mobile: 123456789,987654321 # mobile phone number, multiple phone number joint by `,`
       sign: "xxxxxxxx" # need to register; usually brand name
 ```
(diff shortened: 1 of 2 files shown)


proditis:
Another suggestion could be to include the easeprobe hostname at the footer area but i dont how to get the current hostname :smile: 
image

proditis:
Ok i found how i can get the hostname not sure if this is correct any suggestions are more than welcome
```go
	myname, err := os.Hostname()
	if err != nil {
		myname = "localhost.localdomain"
	}

	discord.Embeds = append(discord.Embeds, Embed{
		Author:      Author{},
		Title:       result.Title(),
		URL:         "",
		Color:       color,
		Description: description,
		Timestamp:   result.StartTime.UTC().Format(time.RFC3339),
		Thumbnail:   Thumbnail{URL: c.Thumbnail},
		Fields:      []Fields{},
		Footer:      Footer{Text: "Probed from " + name + " at", IconURL: global.Icon},
	})
```

haoel:
Hi @proditis ,

Thanks for this PR. And I'd to make this to be general, not just for Discord. So, I think we can add an EaseProbe Name field in the settings configuration.  This is the same for the hostname. 

To be general, there are some works that need to do, let me take over this. 

I will close this PR and recreate another PR.

-- Hao

proditis:
@haoel Sure thing, closing this then.

PR #75, opened 2022-05-11
Create the `easeprobe.pid` file & Accept HUP Signal to Rotate Log

haoel:
two major features:

- **pid file**: we can configure the PID file in settings, and if configured to empty "", it would be disabled. 
   - If the file already exists, EaseProbe would overwrite it. 
   - If the file cannot be written, EaseProbe would exit with an error.
   - If want disables the PID file, you can configure the pid file to "".

- **HUP signal handling**: rotate the access log of the application log once receive the HUP signal. 
  
   ```YAML
    log:
      file: /path/to/log/file
      self_rotate: true # default: true
    ```
  
    If `self_rotate` is `true`, EaseProbe would rotate the log automatically, and the following options are available: 
  
    ```YAML
      size: 10 # max size of log file. default: 10M
      age: 7 # max age days of log file. default: 7 days
      backups: 5 # max backup log files. default: 5
      compress: true # compress. default: true
    ```
     If `self_rotate` is `false`, EaseProbe would not rotate the log, and the log file would be rotated by the 3rd-party tool. e.g. `logrotate`.


cmd/easeprobe/main.go
```diff
@@ -54,6 +56,24 @@ func main() {
 		os.Exit(-1)
 	}
 
+	// Create the pid file
+	d, err := daemon.NewPIDFile(c.Settings.PIDFile)
+	if err != nil {
+		log.Errorf("Error: Cannot create the PID file: %s", err)
+		// Try Working Directory
+		pidfile := filepath.Join(global.GetWorkDir(), global.DefaultPIDFile)
```
proditis, line 64: if i understand this correctly, when we fail to create the pid file defined by the settings, we try to create a pid file on the current directory? 

If so this is kind of counter intuitive. I think that the proper action when we fail to create a pid is to not start.

README.md
```diff
@@ -617,6 +617,9 @@ notify:
 # Global settings for all probes and notifiers.
 settings:
 
+  # Daemon settings
+  pid: /var/run/easeprobe.pid # pid file path, default: $CWD/easeprobe.pid
```
proditis, line 621: There should be an option to disable pid file writing all together if needed. This will make easeprobe easier to run on read-only containers

daemon/daemon_linux.go
```diff
@@ -0,0 +1,32 @@
...
+package daemon
+
+import (
+	"os"
+	"path/filepath"
+	"strconv"
+)
+
+func processExists(pid int) bool {
+	_, err := os.Stat(filepath.Join("/proc", strconv.Itoa(pid)))
```
proditis, line 30: This will stop easeprobe functioning under non linux systems :cry:

daemon/daemon.go
```diff
@@ -0,0 +1,85 @@
...
+func NewPIDFile(pidfile string) (*Config, error) {
+	c := &Config{
+		PIDFile: pidfile,
+	}
+	_, err := c.CheckPIDFile()
+	if err != nil {
+		return nil, err
+	}
+
+	if err := os.MkdirAll(filepath.Dir(c.PIDFile), 0755); err != nil {
```
proditis, line 46: Do we really need to create the entire path tree for creating the pid file? Most daemons rely on the fact that the path at which they save their pid exists and its (usually) the administrators responsibility to make sure it exists.

cmd/easeprobe/main.go
haoel: Yes, your understanding is right!

And thanks for this review, I will correct the behavior as you mentioned - cannot create the pid file, just failed to start.

daemon/daemon_linux.go
haoel: No worry. This file is only compiled for the non-windows and non-mac systems. 

notice the file name with the `_linux` suffix. and there are two comment lines at the beginning of this file to indicate the build platform.

```
//go:build !windows && !darwin
// +build !windows,!darwin
```

proditis: oh ok actualy this is what confused me the `!windows and !darwing` excludes these, so `linux` & `openbsd` should be included when trying to build this?

daemon/daemon.go
haoel: I think it's harmless. If the administrator can guarantee the directory, this won't hurt anything. and some of the other daemons do the same thing, such as - docker.

README.md
haoel: thanks for this input. If we want to disable the pid file, just configure the empty string.

daemon/daemon.go
proditis: :+1: I am generally cautious when applications make too many decisions for me. This helps when faced with unexpected problems. 

Most of the daemons i've seen have this folder creation at their init script (ssh, nginx, apache). 

PS: Sorry if i sound too cautious but i've seen far too many applications fail in insanely unexpected ways :smile:

daemon/daemon.go
```diff
@@ -0,0 +1,89 @@
...
+	}
+
+	c.pidFd, _ = os.OpenFile(c.PIDFile, os.O_APPEND|os.O_EXCL, 0600)
+	return c, nil
+}
+
+// CheckPIDFile check if the pid file exists
+// if the PID file exists, return the PID of the process
+// if the PID file does not exist, return -1
+func (c *Config) CheckPIDFile() (int, error) {
```
proditis, line 66: Suggestion to also check that the pid file is indeed a file and not symlink or whatnot?

daemon/daemon_linux.go
haoel: Yes, in Go land, we can run `go tool dist list` you can see all of the platform and CPU arch it supports to do the cross-compiling. (The list as below)

And we can use the `go:build` or `+build` to let go compiler understand the source file can be compiled on what platform. for the `processExists()` function, I implement 3 version for each different platform. 

```shell
$ go tool dist list                                                                                            ─╯
aix/ppc64
android/386
android/amd64
android/arm
android/arm64
darwin/amd64
darwin/arm64
dragonfly/amd64
freebsd/386
freebsd/amd64
freebsd/arm
freebsd/arm64
illumos/amd64
ios/amd64
ios/arm64
js/wasm
linux/386
linux/amd64
linux/arm
linux/arm64
linux/mips
linux/mips64
linux/mips64le
linux/mipsle
linux/ppc64
linux/ppc64le
linux/riscv64
linux/s390x
netbsd/386
netbsd/amd64
netbsd/arm
netbsd/arm64
openbsd/386
openbsd/amd64
openbsd/arm
openbsd/arm64
openbsd/mips64
plan9/386
plan9/amd64
plan9/arm
solaris/amd64
windows/386
windows/amd64
windows/arm
windows/arm64
```

proditis: Thank you so much for the explanation :heart:

daemon/daemon.go
haoel: Sometimes, we like the code can do as much as it can to make the great availability, but sometimes not.   In this case, something we probably forgot creates the directory, so we can help to create it, this behavior is the same just like the logs and the data files, but sometimes it might not be, because we configured the incorrect directory. 

And I'm a lazy person, I like the code that can do more works.  ;-)

BTW, Your feedback is very important!  Please keep giving such feedback to me. it's valuable.

daemon/daemon.go
haoel: made this change.

haoel:
@proditis @zhao-kun  I've updated this PR's description (at the beginning), please help to review it. Thanks in advance!

conf/log.go
```diff
@@ -105,14 +124,96 @@ func (l *Log) CheckDefault() {
...
+
+//Rotate rotate the log file
+func (l *Log) Rotate() {
+	if l.Writter != nil {
+		if lumberjackLogger, ok := l.Writter.(*lumberjack.Logger); ok {
+			// self rotate
+			if err := lumberjackLogger.Rotate(); err != nil {
+				log.Errorf("[Log] Rotate log file failed: %s", err)
+			}
+		} else if fileLogger, ok := l.Writter.(*os.File); ok {
```
zhao-kun, line 182: The `Writer` could be `os.Stdout`, we should not close it in that case.

cmd/easeprobe/main.go
```diff
@@ -54,6 +56,21 @@ func main() {
...
 	}
 
+	// Create the pid file if the file name is not empty
+	if len(strings.TrimSpace(c.Settings.PIDFile)) > 0 {
+		d, err := daemon.NewPIDFile(c.Settings.PIDFile)
+		if err != nil {
+			log.Fatalf("Fatal: Cannot create the PID file: %s!", err)
+			os.Exit(-1)
+		}
+		log.Infof("Successfully create the PID file: %s", d.PIDFile)
```
proditis, line 66:
```suggestion
		log.Infof("Successfully created the PID file: %s", d.PIDFile)
```

README.md
```diff
@@ -233,6 +234,56 @@ Check the  [Notification Configuration](#37-notification-configuration) to see h
...
+  If `self_rotate` is `true`, EaseProbe would rotate the log automatically, and the following options are available: 
+
+  ```YAML
+    size: 10 # max size of log file. default: 10M
+    age: 7 # max age days of log file. default: 7 days
+    backups: 5 # max backup log files. default: 5
+    compress: true # compress. default: true
+  ```
+
+  If `self_rotate` is `false`, EaseProbe would not rotate the log, and the log file would be rotated by the 3rd-party tool. e.g. `logrotate`.
```
proditis, line 282: and the log file would have to be rotated by a 3rd-party tool (such as `logrotate`) or manualy by the administrator
```shell
mv /path/to/easeprobe.log /path/to/easeprobe.log.0 
kill -HUP `cat /path/to/easeprobe.pid`
```

cmd/easeprobe/main.go
```diff
@@ -54,6 +56,21 @@ func main() {
...
+		defer d.RemovePIDFile()
+	} else {
+		log.Info("No PID file is created.")
```
proditis, line 69:
```suggestion
		log.Info("Skipping PID file creation (pidfile empty).")
```

daemon/daemon_darwin.go
```diff
@@ -0,0 +1,30 @@
...
+ */
+
+package daemon
+
+import "golang.org/x/sys/unix"
+
+func processExists(pid int) bool {
+	// OS X does not have a proc filesystem.
+	// Use kill -0 pid to judge if the process exists.
+	err := unix.Kill(pid, 0)
```
proditis, line 28: Do we keep this check after all for all systems?

conf/log.go
haoel: this check makes sure the Writer is `*os.File` type only.

proditis (commented):
Everything looks good to me (up to the point I can understand).

Great work everyone :clap:

cmd/easeprobe/main.go
proditis: I couldnt find a better way to word this, but i think its good if we mention the reason somehow (could be an AHA! moment for someone)

proditis (approved):
:100:

daemon/daemon_darwin.go
haoel: This check is only for Unit-Test now(make sure we write the right pid), it has nothing to do with the production code. On production, we won't check this anymore.

README.md
haoel: The document updated!

README.md
```diff
@@ -279,7 +279,12 @@ There are some administration configuration options:
     compress: true # compress. default: true
   ```
 
-  If `self_rotate` is `false`, EaseProbe would not rotate the log, and the log file would be rotated by the 3rd-party tool. e.g. `logrotate`.
+  If `self_rotate` is `false`, EaseProbe would not rotate the log, and the log file would be rotated by the 3rd-party tool (such as `logrotate`) or manualy by the administrator.
```
proditis, line 282:
```suggestion
  If `self_rotate` is `false`, EaseProbe would not rotate the log, and the log file would be rotated by the 3rd-party tool (such as `logrotate`) or manually by the administrator. 
```

proditis (approved):
:+1:

conf/log.go
zhao-kun: Type of https://pkg.go.dev/os#Stdout look like *os.File

zhao-kun (approved):

PR #78, opened 2022-05-12
Initial roadmap draft

proditis:
This is the initial draft for the EaseProbe Roadmap.

It needs further cleanup and a few directions from the rest of the team with regards to some of the things i've added.

Such as: 
> * Add support for host group probes (eg 1 host definition with 4 services) 

I would like to discuss these as I am not strongly for them but still came across the need for them when i was working with easeprobe, but i can find ways to work around that need for now. 

Also some items are more like "feature requests" rather than roadmap entries, please provide advise if you want me to open issues for them and link the issue numbers so that it will look closer to the format that can be found eg on the EaseMesh Roadmap https://github.com/megaease/easemesh/blob/main/docs/Roadmap.md.


qicz (approved):

nullsimon:
there are several features that I'm care about

1 support custom probe with different notifies, notifies group or notifies list are both great
2 support metrics and threshold value, absolute value and relative value comparison (need history record support)
3 for notify action, I think we should use message queue to send notifies, another app to deal with notifies, for some service, maybe restart, auto scale, for other service, maybe have to notify some one online to deal with it.

docs/Roadmap.md
```diff
@@ -0,0 +1,112 @@
...
+  - [Roadmap 2022](#roadmap-2022)
+    - [General](#general)
+    - [Probes](#probes)
+    - [Notify](#notify)
+
+
+## Product Principles
+EaseProbe tries to do two things but do them extremely well - **Probe** and **Notify**.
+
+1. **Probe and Notify**. Aims to provide the ability to easily **probe** servers, services and API endpoints reliably and **notify** on status updates.
```
haoel, line 18: I'd like the first principle is lightweight, standalone, and KISS.

docs/Roadmap.md
```diff
@@ -0,0 +1,112 @@
...
+* Add opt-out options where appropriate
+  * [x] megaease/easeprobe#75 add opt-out option for `logfile` option 
+  * Add opt-out option for SLA data persistence `data: false`
+  * Make historical data configurable `history: false` and avoid creating backups of statistics
+
+### Probes
+* integrate automatic service discovery which includes probing details
+* add plain old icmp ping probe (or read next)
+* shell probe command must allow allocation or not of `stdout`/`stderr` or redirection of them (eg right now we are forced to create wrapper scripts to achieve this) This will also make ping by shell usable
+* Add support to define desired notification on probes???
```
haoel, line 74: how about introducing a group concept? something like:

```yaml
GroupA:  
   http:
   server:
   client:
   notify:
      discord:
      slack:
GroupB:
   ....
```

docs/Roadmap.md
```diff
@@ -0,0 +1,112 @@
...
+## Roadmap 2022
+Some of the features that are planned/considered for 2022 can be broken down into three categories: *General*, *Probe*, *Notify*
+
+
+### General
+* Work on detailed documentation
+* support for common daemon features
+  * [x] megaease/easeprobe#75 add `daemon()` & `/var/run/easeprobe.pid` support
+  * [x] megaease/easeprobe#75 add SIGHUP, and ensure it closes and re-opens of logfile to allow for `easeprobe.log` rotation
+  * add syslog support eg `logfile: syslog`
```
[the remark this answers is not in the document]
proditis: Basically whatever goes to easeprobe.log must have an option to be send on syslog if needed. At least that was the idea, for us to use existing logging infrastructure that there might be in place.

docs/Roadmap.md
proditis: This could work i think, the functionality i was aiming was more on the lines of:
i have two probes, but one of them needs priority in resolving any issues. For that reason i configure say SMS notify, but for the other i dont care that much so i want it to go to the general chat of a discord server.

With the grouping you describe we will end up with two groups with 1 notification each (?).

docs/Roadmap.md
```diff
@@ -0,0 +1,112 @@
...
+* add plain old icmp ping probe (or read next)
+* shell probe command must allow allocation or not of `stdout`/`stderr` or redirection of them (eg right now we are forced to create wrapper scripts to achieve this) This will also make ping by shell usable
+* Add support to define desired notification on probes???
+```yaml
+tcp:
+  - name: Memcached
+    host: 10.7.0.253:11211
+    notify: "Server #Alert" # notify.discord.name
+```
+* Add support for host group probes (eg 1 host definition with 4 services)
```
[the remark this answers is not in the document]
proditis: Terribly sorry for leaving these details out, i was having a hard time to describe it. 

So the idea is that a service can have more than one systems serving it such as a cluster of load balanced webservers. The actual SLA and certain reports should not address the individual "systems" but rather the overall performance of the web service across the servers. 

Alternative: 1 server with 4 different services, all working under the name of `MyServer`, when i get reports about problems i still get to see the individual failures but my daily reports will now have these 4 different services under a single grouping.

I hope this clears up what i had in mind here.

docs/Roadmap.md
proditis: Added please have a look

proditis:
> there are several features that I'm care about
> 
Awesome, thank you for your feedback @nullsimon 

> 1 support custom probe with different notifies, notifies group or notifies list are both great 

So just to make sure i understand, this refers to notification groups instead of addressing them individually? eg
```yaml
http:
  # Some of the Software support the HTTP Query
  - name: ElasticSearch
    url: http://elasticsearch.server:9200
    notify: GroupA
  - name: Prometheus
    url: http://prometheus:9090/graph
    notify: GroupB
groupA:
 discord:
 slack:
groupB:
 discord:
 slack:
```

I like this as it covers more use-cases than what I initially described and is way more flexible.

> 2 support metrics and threshold value, absolute value ~relative value comparison (need history record support, too much for a probe app :), we can just use another app to deal with it)~ 

This one already included or referring to something new?

> 3 for notify action, I think we should use message queue to send notifies, another app to deal with notifies, for some service, maybe restart, auto scale, for other service, maybe have to notify some one online to deal with it.

I think the current notify logic can also cover this operation. I was thinking about it and there is no major change for easeprobe if the "notification" goes to discord for a notification purpose or a message queue for workers to recover a failed service. 

What the receiver of a notification does should not be the concern of easeprobe. The actual service recovery, i agree with you, MUST be another tool and not our job to do.

nullsimon:
> ```yaml
> http:
>   # Some of the Software support the HTTP Query
>   - name: ElasticSearch
>     url: http://elasticsearch.server:9200
>     notify: GroupA
>   - name: Prometheus
>     url: http://prometheus:9090/graph
>     notify: GroupB
> groupA:
>  discord:
>  slack:
> groupB:
>  discord:
>  slack:
> ```

```yaml
http:
  # Some of the Software support the HTTP Query
  - name: ElasticSearch
    url: http://elasticsearch.server:9200
    notify: GroupA,telegram
  - name: Prometheus
    url: http://prometheus:9090/graph
    notify: telegram,discord
groupA:
 discord:
 slack:
 GroupB:
groupB:
 discord:
 slack:
```
just like this, support notifies list, and also support notify group abstraction

docs/Roadmap.md
haoel: Thanks for the clarification, it makes sense to me. 

And yes, your understanding is right - two groups with 1 notification for each. **Basically, the entity-relationship is** -  **the probes are grouped by notfication.** 

And,  to cover all of the scenarios, the best model is the `n:m` relationship between probe and notification,  but this would introduce the complexities of configuration - we need an extra configuration to describe the relationship.

something like.

```yaml
relationsip:
- probe_id_1 : notification_id_1
- probe_id_2 : notification_id_1
- probe_id_1 : notification_id_2
....
```

haoel:
I think we need a dedicated thread to discuss the list, group, and the relationship, this is quite important. In this Roadmap, we can just state the direction, problems, and issues. 

Thanks for your efforts @proditis

docs/Roadmap.md
proditis: Yes i was facing a similar problem when i tried to _describe_ it with some fake configuration examples just to see what made more sense but it gets confusing very easily. 

The current configuration layout for easeprobe is a beauty i would hate to make it confusing just to support these features, but maybe we can start with a simple support for it and expand it later as we get more data and use-cases? 

stage1: ability to point to a single notify end point based on the current name
stage2: ability to support more elaborate groupings
stage3: full relationships support between probes and notify endpoints

Just an idea

docs/Roadmap.md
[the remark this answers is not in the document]
haoel: it's quite clear. And I could imagine some statistical analysis works, and it could have other similar scenarios.  So, I think we need a general abstraction solution here like a math formula. Keep thinking hard ...

haoel (approved):

PR #79, opened 2022-05-13
Support the customized discord name and icon

haoel:
This is the work for #73


proditis (approved):
Looks good to me . I love the idea of the EaseProbe object :smile:

Issue #69, opened 2022-05-05
[feature request]Hot reloading support


douglarek:
**Background & X Problem**
Currently (2022-05-05), easeprobe does not support hot reloading, so the problem is that if I just modify the config.yaml file (which can be a frequent action), then I need to restart the application for it to take effect.

**Proposal and Expectation**
We can add a reload option to the configuration file to enable it:

```yaml
reload:
  enabled: true # default is false
  period: 10s # scan interval, default is 10s
```

**Solutions & Y Problems**
We need to not lose the SLA information that the configuration has not changed when the reload is turned on.

proditis:
Hi @douglarek and @haoel I may have an alternative suggestion for this, not sure if this also satisfies your original requirements, but it may be better if we did this through a signal handler (eg SIGHUP). It will also be much simpler to implement (i think)?

Having the application reread the configuration all the time without an admin intervention is a bit dangerous imho. I much rather have control of when the application re-reads its configuration.

What do you think?

haoel:
Actually, I thought this feature a couple of weeks ago.

To accomplish this feature, We need to do understand how we change the prober's configuration and keep the SLA statistics. 

There are several works need to do

0) Detect the configuration file changes or just handle a signal likes @proditis mentioned above.

1) We need a uniq identity for each prober, so that we can do add, delete, and update operations. Maybe we can reuse the name of prober to do this(back compatibility)

2) We can put the SLA statistics into a dedicated map[probe-name, SLA]. Then, we can simply destroy all of probes and recreate all probers. This need a bit more refactoring. 

For notifications configuration changes, just destroy and recreate all of them.

But we need to know the SLA is still stored in memory, I am thinking about persistent them, if we persistent that, we can free to restart the EaseProbe.

I quite like to persistent the SLA statistics into a file. How do you guys think?

- Hao

haoel:
@douglarek I think the X-problem of this feature should be that keep the SLA statistics after the process restart, and the Y-Problem is reloading the configuration without exiting the process. Right?

-- Hao

proditis:
@haoel I agree, having the SLA statistics persist into a file will solve more than just this problem :smiley: (such as restarts for updating the actual easeprobe binary)

douglarek:
> @douglarek I think the X-problem of this feature should that keep the SLA statistics after the process restart, and the Y-Problem is reloading the configuration without exiting the process.
> 
> Am I Right? ;-)
> 
> -- Hao

totally agree.

[...]

PR #81, opened 2022-05-14
Support max backups of SLA data files

haoel:
Support to configure the max of SLA data file backups. default: 5
If set to a negative value, keep all backup files

```yaml
  sla:
    data: /tmp/data/data.yaml
    backups: 20
```

keep addressing the issue  #69

### DIFF
--- a/probe/data.go
+++ b/probe/data.go
@@ -21,8 +21,10 @@ import (
 	"io/ioutil"
 	"os"
 	"path/filepath"
+	"sort"
 	"time"
 
+	log "github.com/sirupsen/logrus"
 	"gopkg.in/yaml.v3"
 )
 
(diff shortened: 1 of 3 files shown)


probe/data.go
```diff
@@ -100,3 +102,36 @@ func LoadDataFromFile(filename string) error {
...
+		return
+	}
+
+	// if backups is not exceed the max number of backup files, return
+	if len(matches) <= backups {
+		log.Debugf("No need to clean data file (%d - %d) ", backups, len(matches))
+	}
+
+	// remove the oldest backup files
+	sort.Strings(matches)
```
proditis, line 128: I am not sure how this sorting works, but usually numeric sorting of strings has a few issues, check the example with linux sort command, make sure we dont end up removing a newer file by accident.
```shell
$ seq 1 10|sort
1
10
2
3
4
5
6
7
8
9
```

proditis (commented):
Looks good to me

Great work @haoel as always

probe/data.go
haoel: The backup files end with the time RFC3339 format string. So, I think the string sort should be fine. 

```
data.yaml-2022-05-13T17:10:35
data.yaml-2022-05-13T17:13:08 
data.yaml-2022-05-14T15:28:08 
data.yaml-2022-05-14T15:30:39 
data.yaml-2022-05-14T15:34:35
```

probe/data.go
```diff
@@ -95,8 +97,41 @@ func LoadDataFromFile(filename string) error {
...
+
+	for i := 0; i < len(matches)-backups; i++ {
```
zhao-kun, line 130: If I understand correctly the negative value will lead to removing all backup files? The behavior is deviating from the description of the PR.

haoel: At the beginning of this function, we checked the negative value. it would keep all backup files.

```go
	// if backups is negative value, keep all backup files
	if backups < 0 {
		return
	}
```

zhao-kun (approved):

PR #84, opened 2022-05-15
Minor Dockerfile/entrypoint.sh edit and documentation updates

proditis:
* minor changes from `would` to `will` readme
* added mention on README for docker image use and release download
* added environment variable on Dockerfile `PROBE_CONFIG`
* Updated entrypoint.sh to set default value if not set for `PROBE_CONFIG`
* Removed `-f /opt/config.yaml` from default entrypoint action

### DIFF
--- a/README.md
+++ b/README.md
@@ -279,7 +279,7 @@ There are some administration configuration options:
     compress: true # compress. default: true
   ```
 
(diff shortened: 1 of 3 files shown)


haoel (approved):
Thank for this PR!

PR #86, opened 2022-05-15
bug-fix: correct the `$PROBE_CONFIG` assignment in entrypoint.sh

haoel:


resources/scripts/entrypoint.sh
```diff
@@ -4,7 +4,7 @@
 # 1) run easeprobe without any arguments
 # 2) run easeprobe with easeprobe arguments
 # 3) run the command in easeprobe container
-${PROBE_CONFIG:=/opt/config.yaml}
+PROBE_CONFIG=${PROBE_CONFIG:=/opt/config.yaml}
```
proditis, line 7: Missed the error when i was testing locally 

```suggestion
PROBE_CONFIG=${PROBE_CONFIG:-/opt/config.yaml}
```

haoel: me too. ：-）

And this change let me google differences -  https://stackoverflow.com/questions/2013547/assigning-default-values-to-shell-variables-with-a-single-command-in-bash

Good!

proditis (approved):

PR #87, opened 2022-05-16
fix undefined processExists error

proditis:
just a copy of daemon_darwin to fix daemon/daemon.go:90:5: undefined: processExists

I tried changing the headers to make the daemon_darwin be used but it kept failing. 
```
//go:build darwin && openbsd
// +build darwin,openbsd
```
Is this command i'm using to cross-compile easeprobe wrong?
```shell
GOOS=openbsd GOARCH=amd64 go build -a -ldflags '-s -w -extldflags "-static"' -gcflags=-G=3 -o $PWD/build/bin/easeprobe.obsd $PWD/cmd/easeprobe

```


haoel:
This is a very good finding. And I realize I made a mistake - the file `daemon_linux.go` should be `daemon_unix.go`

- the suffix `_linux` only for the Linux platform
- the suffix `_darwin` only for macOS
- the suffix `_unix` is for all Unix-like platform 

So, I think the correct fix as below
1)  rename the `daemon_linux.go` to `daemon_unix.go`, I believe the OpenBSD and other Unix-like platforms can be compiled successfully.

2) evaluating the `daemon_unix.go` whether works on the OpenBSD platform, if not, we need the `daemon_openbsd.go`, and exclude the `openbsd` in `daemon_unix.go`

haoel:
Just Google how to test it. we can run the following command to check how the `go build` choose the file.

```
cd daemon
 GOOS=openbsd GOARCH=amd64 go list -f '{{.GoFiles}}'
```

proditis:
> * rename the `daemon_linux.go` to `daemon_unix.go`, I believe the OpenBSD and other Unix-like platforms can be compiled successfully.

Linux is not a good candidate for BSD systems since these have no `/proc` (mostly). In this case the best candidate is darwin as it uses closer to BSD API's.

I see two options
1. rename the existing _openbsd to _unix and have a small duplication of code? 
2. rename the _darwin => _unix and remove _openbsd?

What do you think?

haoel:
I like the 2nd one, as darwin is based on the BSD system. 

And don't forget to remove the go-build tags in `_linux.go`,  and add the correct tags in `_unix.go`

proditis:
The list of files each os pulls
```
linux: [daemon.go daemon_linux.go]
windows: [daemon.go daemon_windows.go]
darwin: [daemon.go daemon_unix.go]
openbsd: [daemon.go daemon_unix.go]
```

daemon/daemon_unix.go
```diff
@@ -1,5 +1,5 @@
-//go:build darwin
-// +build darwin
+//go:build darwin || openbsd
+// +build darwin openbsd
```
haoel, line 2: how about adding freebsd and netbsd?

proditis: :man_facepalming: How could i ever ?!?!?! :smile: adding the rest of the family

proditis: What do you think if we make this our fall back for unix systems and simply exclude linux/windows on top ?

There are a lot more unix systems that will work with this one rather with the linux.

haoel (approved):

daemon/daemon_unix.go
haoel: I like the `!linux && !windows`

haoel:
Wonderful! @proditis

Issue #89, opened 2022-05-16
HTTP server bind already in use should cause the daemon to exit


proditis:
When the port is already in used, easeprobe displays the following error but continues to run. 
```
ERRO[2022-05-16T00:56:00Z] [Web] HTTP server error: listen tcp 127.0.0.1:8181: bind: address already in use 
```

I believe that it would be better if we exited with a non zero status code when this happens.

proditis:
I was thinking something like this 
```go
		if err := http.ListenAndServe(host+":"+port, r); err != nil {
			log.Errorf("[Web] HTTP server error: %s", err)
			os.Exit(-1)
		}
```

haoel:
Yes, you find the correct place. 

However, the HTTP server is running on a go-routine,  we need a mechanism to notify the main thread the HTTP server failed and the main thread can be exited! 

I am thinking the solution now...

proditis:
Confirmed fixed :+1:

PR #90, opened 2022-05-16
EaseProbe exits if the web server address already in use

haoel:
This PR addresses the issue #89 

1)  splite the `http.LisentAndServe()` to  `net.Lisen()` and `http.Serve()`
2) move the `web.Server()` before starting the probe and notifcation, so that easeprobe can be exited safely.


zhao-kun (approved):
LGTM

PR #92, opened 2022-05-16
Disable SLA persistence when conf.Settings.SLAReport.DataFile is "-"

proditis:
This PR does
* [x] removes the `Join(dir,'data',file)` so that our `data.yaml` file location is predictable
* [x] Changes the decision tree logic for the `data.yaml` file based on the specifications agreed with @haoel 
* [x] disables logging when conf.Settings.SLAReport.DataFile is explicitly set to `"-"`
* [x] updates SLA data Persistence section to document the behavior of `data: "-"`
* [x] merge a duplicate readme notice
* [x] Update global.DefaultDataFile to `data/data.yaml` so that by default we use a sub-folder for the data files

Cases 
* Case One: No data file is configured, then we use the $CWD/data.yaml as the default.
* Case Two: data file name configured, but not configured the directory, such as: `data: mydatafile.yaml`, we use the `$CWD/mydatafile.yaml`
* Case Three: the full path has been configured. we just simply use the full path.
* Case Four: the disable flag configured `-` eg `data : -` , then we disable the data file.

### DIFF
--- a/README.md
+++ b/README.md
@@ -37,9 +37,10 @@ EaseProbe would do three kinds of work - **Probe**, **Notify**, and **Report**.
 
 Ease Probe supports the following probing methods: **HTTP**, **TCP**, **Shell Command**, **SSH Command**,  **Host Resource Usage**, and **Native Client**.
 
-> **Notes**:
+
+> **Note**:
 >
-> The prober name is a unique ID, DO NOT use the same name for a different prober.
+> **Keep in mind that the prober name must be **unique** among probes. If multiple probes are defined with the same name, it could lead to corruption of the metrics data and the behavior of the application will be non-deterministic.**
 
 - **HTTP**. Checking the HTTP status code, Support mTLS, HTTP Basic Auth, and can set the Request Header/Body. ( [HTTP Probe Configuration](#31-http-probe-configuration) )
 
@@ -183,7 +184,7 @@ notify:
   sms:
     - name: "sms alert service"
       provider: "yunpian"
-      key: xxxxxxxxxxxx # yunpian apikey 
+      key: xxxxxxxxxxxx # yunpian apikey
       mobile: 123456789,987654321 # mobile phone number, multiple phone number joint by `,`
       sign: "xxxxxxxx" # need to register; usually brand name
 ```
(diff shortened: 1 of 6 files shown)


haoel:
I'd like the SLA persistent data as the default behavior, this could reduce the configuration.  This PR changes the default behavior to disable the SLA data. So, I suggest using the explicit configuration to disable it. For example: `data :"-"`.

I still like the `data` directory to be the default because the would guide the people to the best practice.

haoel:
as we discussed in issue #69. 

### Case Study

- **Case One**: No data file configured, then we use the `$CWD/data/date.yaml` as the default.
- **Case Two**: data file name configured, but not configured the directory, such as: `mydatafile.yaml`, we use the `$CWD/data/mydatafile.yaml`
- **Case Three**: the full path has been configured. we just simply use the full path.
- **Case Four**: the disable flag configured - `data : - `, then we disable the data file.


### Pseudo Code 

So, the code logic should be:

```
if datafile == `-` then return
dir, file := filepath.split(datafile)
if dir == "" then dir = filepath.join( getcwd(), "data")
if file == "" then file = `data.yaml`

filepath.join(dir, file)
```

proditis:
I'm going to need some help with the windows test, i dont have access to windows and have no idea how to troubleshoot this error.  I got beaten by the tests :smile:

haoel:
Thanks for your great work. 

Just one comment - we also need to take the function - `saveData()` in `main.go`. It would output debug messages even if we disable the data file.

The simple fix just defines the internal `save` function to empty. 

```go
save := func() {
        ....
}
if c.Settings.SLAReport.DataFile != "-" {
	save = func() {}
}
```


By the way, for the Windows test environment, the current easiest way is to write the unit test, the Github Action will run it on the windows platform. The integration test would be complicated, I think your current test scripts might not be suitable for Windows...

proditis:
I think i finally made them work. I examined how the test was executed prior to my changes and adapted it accordingly, now its almost identical to the prior test.

The problem was hidden by the fact we were MkdirAll() inside SaveDataToFile before so the hard-coded path for /tmp/easeprobe was created by us. I tried to play "smart" and update it to use TmpDir and TmpFile but under windows they produced some unexpected behavior (still dont know what exactly happened that caused to fail the test).

I think its time to create a small windows vm to run the tests locally and not fill up the commit history

haoel (approved):
Thank for this PR.  Wonderful!

PR #95, opened 2022-05-18
feat: Support to export the metrics to Prometheus

haoel:
This is the first version to export the metrics to promethues.

- [x] endpoint: `https://localhost:8181/metrices/`
- [x] metrics naming: `[easeprobe name]_[probe kind]_[tag]_[metric]`.  If the `tag` is empty, just ignore. For examples: `easeprobe_http_total`, `easeprobe_host_cpu`, `easeprobe_client_redis_duration` , `easeprobe_shell_exit_code` etc. 
- [x] probe metrics: 
    -  [x] counter: total [name, status] - number of probe
    - [x] gauge duration [name, status] - probe duration
    - [x] gauge: status [name] - probe status
    - [x] gruge: sla[name] - probe sla 
 - [x] host metrics:
    - [x] gauge: cpu [host, state]
    - [x] gauge: memory [host, state]
    - [x] gauge: disk [host, state]
- [x] http metrics 
    - [x] counter: status_code[name, status] 
    - [x] gauge: content_len[name, status] 
- [x] shell metrices 
    - [x] counter: exit_code[name, exit ] 
    - [x] gauge: output_len[name, exit]
- [x] ssh metrics 
    - [x] counter: exit_code[name, exit ] 
    - [x] gauge: output_len[name, exit]

The snapshot below is the Grafana panel for host CPU metrics 

image


probe/base/base.go
```diff
@@ -123,12 +129,34 @@ func (d *DefaultOptions) Probe() probe.Result {
...
+	d.metrics.Total.With(prometheus.Labels{
+		"probe":  d.ProbeName,
+		"status": d.ProbeResult.Status.String(),
+	}).Inc()
+
+	d.metrics.Duration.With(prometheus.Labels{
+		"probe": d.ProbeName,
+	}).Set(d.ProbeResult.RoundTripTime.Seconds())
+
+	status := 1 // up
```
proditis, line 151: minor beatification question but can we have something like `#define UP 1` (i dont know if an equivalent in go exists) so that we can do `status := UP // 1`  and `status = DOWN` ?

metric/prometheus.go
```diff
@@ -0,0 +1,125 @@
...
+)
+
+const module = "Metric"
+
+// MetricsType is the generic type of metrics
+type MetricsType interface {
+	*prometheus.CounterVec | *prometheus.GaugeVec | *prometheus.HistogramVec | *prometheus.SummaryVec
+}
+
+var counterMap = make(map[string]*prometheus.CounterVec)
```
jordy1024, line 32: four `maps define` include `four types` and belong to a set of definitions. 
For readability, it is more recommended to:
```
var (
    counterMap   = make(map[string]*prometheus.CounterVec)
    gaugeMap     = make(map[string]*prometheus.GaugeVec)
    histogramMap = make(map[string]*prometheus.HistogramVec)
    summaryMap   = make(map[string]*prometheus.SummaryVec)
)
```

probe/base/base.go
```diff
@@ -23,10 +23,18 @@ import (
...
 	"github.com/megaease/easeprobe/probe"
+	"github.com/megaease/easeprobe/report"
+	"github.com/prometheus/client_golang/prometheus"
 
 	log "github.com/sirupsen/logrus"
 )
 
+// Probe Simple Status
+const (
+	UP   int = 1
```
jordy1024, line 34: `Up` and `Down` are better?

jordy1024 (commented):
hello, I have made some small suggestions, probably mostly code readability, which also involves some logic issues.

probe/base/metrics.go
```diff
@@ -0,0 +1,47 @@
...
+package base
+
+import (
+	"github.com/megaease/easeprobe/global"
+	"github.com/megaease/easeprobe/metric"
+	"github.com/prometheus/client_golang/prometheus"
+)
+
+// Metrics is the probe metrics
+type Metrics struct {
```
zhao-kun, line 27: Metrics struct just be referred in the package, so rename to `metrics` is better? Name of the fields should be renamed too?

probe/http/http.go
```diff
@@ -155,6 +160,7 @@ func (h *HTTP) DoProbe() (bool, string) {
 		return false, fmt.Sprintf("HTTP Status Code is %d. It missed in %v", resp.StatusCode, h.SuccessCode)
 	}
 
+	h.ExportMetrics(resp)
```
zhao-kun, line 163: Question: Do we just export metrics for the valid situation?

metric/prometheus.go
```diff
@@ -0,0 +1,125 @@
...
+		name = name[:len(name)-1]
+	}
+
+	log.Debugf("[%s] get the name: %s", module, name)
+	return name
+}
+
+// ValidName check if the char is valid
+func ValidName(ch byte) bool {
+	return (ch >= 'a' && ch <= 'z') || (ch >= 'A' && ch <= 'Z') || (ch >= '0' && ch <= '9') || ch == '_' || ch == '-'
```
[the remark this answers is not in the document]
haoel: Thanks, I am going to refactor the code, and using regex to do the validation.

probe/base/metrics.go
haoel: I think renaming the `Metrics` =>`metrics` and  `NewMetric()` => `newMetrics()`should be fine.

probe/http/http.go
haoel: we need to consider the failed situation.

metric/prometheus.go
```diff
@@ -107,19 +143,40 @@ func GetName(fields ...string) string {
...
+
+// ValidLabelName check if the label name is valid
+func ValidLabelName(label string) bool {
+	return validLabel.MatchString(label)
+}
+
+// ValidMetricChar check if the char is valid for metric name
+func ValidMetricChar(ch byte) bool {
+	return (ch >= 'a' && ch <= 'z') || (ch >= 'A' && ch <= 'Z') ||
+		(ch >= '0' && ch <= '9') || ch == '_' || ch == '-' || ch == ':'
```
jordy1024, line 159: char: `ch == '-'` is valid ?

probe/shell/shell.go
```diff
@@ -38,6 +39,11 @@ type Shell struct {
 	Env                 []string `yaml:"env,omitempty"`
 	Contain             string   `yaml:"contain,omitempty"`
 	NotContain          string   `yaml:"not_contain,omitempty"`
+
+	ExitCode  int `yaml:"-"`
```
zhao-kun, line 43: The inner fields prefer to start with lower-case.

probe/shell/shell.go
```diff
@@ -38,6 +39,11 @@ type Shell struct {
...
+	OutputLen int `yaml:"-"`
```
zhao-kun, line 44: ditto

probe/ssh/ssh.go
```diff
@@ -44,6 +45,11 @@ type Server struct {
 
 	BastionID string    `yaml:"bastion"`
 	bastion   *Endpoint `yaml:"-"`
+
+	ExitCode  int `yaml:"-"`
+	OutputLen int `yaml:"-"`
```
zhao-kun, line 50: ditto

metric/prometheus.go
```diff
@@ -107,19 +143,40 @@ func GetName(fields ...string) string {
...
 
 // RemoveInvalidChars remove invalid chars
 func RemoveInvalidChars(name string) string {
 	var result []byte
-	for i := 0; i < len(name); i++ {
-		if ValidName(name[i]) {
-			result = append(result, name[i])
+	i := 0
+
+	// skip all of the non-alphabetic chars
```
jordy1024, line 167: what is the purpose of the `skip for` code segment ?

metric/prometheus.go
```diff
@@ -90,12 +109,29 @@ func NewGauge(namespace, subsystem, name, metric string,
 	return gaugeMap[metricName]
 }
 
+func getAndValid(namespace, subsystem, name, metric string, labels []string) (string, error) {
+	metricName := GetName(namespace, subsystem, name, metric)
+	if ValidMetricName(metricName) == false {
```
jordy1024, line 114: `ValidMetricName` name means return true if it is a valid name; prefer:
```
if !ValidMetricName(metricName) {
   
}
```

metric/prometheus.go
```diff
@@ -90,12 +109,29 @@ func NewGauge(namespace, subsystem, name, metric string,
...
+
+		return "", fmt.Errorf("Invalid metric name: %s", metricName)
+	}
+
+	for _, l := range labels {
+		if ValidLabelName(l) == false {
```
jordy1024, line 120: ditto

metric/prometheus.go
haoel: Sorry, cannot accept this comment, because this is not easy for reading.

metric/prometheus.go
haoel: because the leading number is invalid!

metric/prometheus.go
haoel: good catch, I will fix the problem.

zhao-kun (approved):

PR #97, opened 2022-05-19
add two conf/conf.go test cases and update isExternalURL

proditis:
This PR adds test cases for `conf/conf.go` and changes the detection method of isExternalUrl 
* [x] conf/isExternalURL
* [x] conf/getYamlFileFromFile

edit: _I changed the scope of the PR to only include 1 module at a time, so that fixes for the test cases can also be included as pair_

### DIFF
--- a/conf/conf.go
+++ b/conf/conf.go
@@ -20,6 +20,7 @@ package conf
 import (
 	"io/ioutil"
 	httpClient "net/http"
+	netUrl "net/url"
 	"os"
 	"path/filepath"
 	"reflect"
@@ -135,8 +136,18 @@ type Conf struct {
 	Settings Settings        `yaml:"settings"`
 }
 
+// Check if string is a url
 func isExternalURL(url string) bool {
-	return strings.HasPrefix(url, "http://") || strings.HasPrefix(url, "https://")
+	if _, err := netUrl.ParseRequestURI(url); err != nil {
+		return false
+	}
+
+	parts, err := netUrl.Parse(url)
+	if err != nil || parts.Host == "" || !strings.HasPrefix(parts.Scheme, "http") {
+		return false
+	}
+
+	return true
 }
 
 func getYamlFileFromInternet(url string) ([]byte, error) {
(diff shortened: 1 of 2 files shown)


proditis:
@haoel i need your guidance with regards to the test cases. What is our policy for when our test cases expose bugs that make them fail? 

I was trying to write some simple tests for functions that seemed easy for me to get start with and one of the cases uncovered this
```
?   	github.com/megaease/easeprobe/cmd/easeprobe	[no test files]
--- FAIL: TestPathAndURL (0.00s)
    conf_test.go:26: isExternalURL("http://") = true, expected false
FAIL
```

I can remove this test for now since its extremely low in severity, but how do you suggest we deal with similar situations in the future? I suspect the more test cases we add the more minor issues like this we will uncover.

haoel:
@proditis I think we should use the better way to check the URL, the url.ParseRequestURI should be the good way to go.

haoel:
> @haoel i need your guidance with regards to the test cases. What is our policy for when our test cases expose bugs that make them fail?

I think if the test case is meaningful, then we need to fix the code, otherwise, we can treat it as low priority or enhancement,  remove the test case and file an issue to track it.

conf/conf.go
```diff
@@ -135,8 +136,18 @@ type Conf struct {
...
+	if _, err := netUrl.ParseRequestURI(url); err != nil {
+		return false
+	}
+
+	parts, err := netUrl.Parse(url)
+	if err != nil || parts.Host == "" || !strings.HasPrefix(parts.Scheme, "http") {
+		return false
+	}
+
+	return true
```
haoel, line 150: I think we can add some logs here for error, info, and debug.

proditis: hmm let me see what I can do.

proditis: Answering my own question: yes we are initialized...

haoel: if not, the StdOut would be the place for log.

haoel (approved):

PR #100, opened 2022-05-20
Versionizing the binary, config, data file

### LINKED ISSUE #98: Add version option for executable and configuration files
**Background & X Problem**
I wanted to performing updates between v1.4.0 and the latest version from repo on my systems, as part of the process I wanted to confirm the version of the binary but we currently have no command line option that will simply display the version. 
As it currently stands those who are interested in tracking the version somehow, have to go through checksums to the binary which is not optimal imho.

Similarly I could not determine for what version of binary my config was for.

**Proposal and Expectation**
I propose the following, 

* addition of a command line option `-v` or `-V` that will just display the version of the current binary
* addition of the version string to the displayed `Usage`
* addition of a `config.yaml` top key `version:` that will allow us to determine what version of binary this config works with
* (optionally) include a `version` key also on the `data.yaml` so that future binaries can detect older versions of the data and accommodate any potential changes that need to take place

_edit: If anyone wants to take on this please be my guest_

haoel:
fixes #98 

- [x] easeprobe -v
```shell
$ ./easeprobe -v                                                                                                                                  ─╯
EaseProbe v1.5.0
```
- [x] easeprobe -h

```shell
$ ./easeprobe -h                                                                                                                                  ─╯
Usage of build/bin/easeprobe:
  -d    dry notification mode
  -f string
        configuration file (default "config.yaml")
  -v    prints version
```

- [x] config.yaml

```yaml
version : 1.5.0
```

- [x] data.yaml 

```yaml
---
name: EaseProbe
version: v1.5.0
---
Probe:
   ...
  ....
```

### DIFF
--- a/README.md
+++ b/README.md
@@ -291,7 +291,7 @@ There are some administration configuration options:
 You can get started with EaseProbe, by any of the following methods:
 * download the release for your platform from https://github.com/megaease/easeprobe/releases
 * use the available EaseProbe docker image `docker run -it megaease/easeprobe`
-* build easeprobe from sources
+* build `easeprobe` from sources
 
 ### 2.1 Build
 
@@ -319,6 +319,7 @@ $ build/bin/easeprobe -f config.yaml
 * `-d` dry run. Can also be achieved by setting the environment variable `PROBE_DRY`
 
 ## 3. Configuration
+
 EaseProbe can be configured by supplying a yaml file or URL to fetch configuration settings from.
 By default EaseProbe will look for its `config.yaml` on the current folder, this can be changed by supplying the `-f` parameter.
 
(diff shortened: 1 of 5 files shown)


proditis:
@haoel Should I be able to see the version metadata inside `data.yaml` ? There was no version information inside.

proditis (approved):
LGTM

haoel:
supplying the unit tests for the data file.

haoel:
Originally, we use the RFC3339  time format as the data file backup file name,  on the Windows platform, it is invalid, because it includes `:`,  in the latest commit, we replace the  `:` with `_` to fix this problem.

proditis:
I think that this was the error i also came across. Wow NICE CATCH :clap:

haoel:
> @haoel Should I be able to see the version metadata inside `data.yaml` ? There was no version information inside.
> 
> 

Yes. 

If the original data.yaml hasn't version data. The new program would write it.

proditis:
@haoel I have tested your pr and when the data.yaml do not exist the meta data are like so
```yaml
---
name: ""
version: ""
---
```

When the file exists and the rotation takes place the new data.yaml has indeed meta information like so
```yaml
---
name: EaseProbe
version: v1.5.0
---
```

I tried track this down but i couldn't, i hope these details help in tracking the case with the empty strings.

haoel:
@proditis  Good catch! I fixed this issue, and add the unit test case in `TestMetaData()` .

proditis (approved):
everything works as advertised :ok_hand:

Issue #104, opened 2022-05-21
Incorporate pprof HTTP debugging handler.


jordy1024:
This registers the pprof debugging handler via the middleware documented in https://pkg.go.dev/github.com/go-chi/chi/middleware#Profiler.

```
r.Mount("/debug", middleware.Profiler())
```

some issues about  `middleware.Profiler()`: https://github.com/go-chi/chi/issues/262

jordy1024:
work in progress

thinhlh:
This is still open
Can not use both `CleanPath` & `Profiler` at the same time
```
router.Use(middleware.CleanPath)
router.Mount("/debug", middleware.Profiler())
```

samanhappy:
Hi @thinhlh, this issue is quite outdated, and the codes referenced have been altered. Could you please detail the problem you encountered? Additionally, it would be helpful if you could provide some context, as both the code and the team have likely seen some changes.

PR #105, opened 2022-05-21
Incorporate pprof HTTP debugging handler(#104)

jordy1024:
support pprof debug tool.

### DIFF
--- a/web/server.go
+++ b/web/server.go
@@ -108,8 +108,7 @@ func Server() {
 	r.Use(middleware.RealIP)
 	r.Use(middleware.Logger)
 	r.Use(middleware.Recoverer)
-	r.Use(middleware.RedirectSlashes)
-	r.Use(middleware.StripSlashes)
+	r.Mount("/debug", middleware.Profiler())
 
 	r.Get("/", slaHTML)
 
--- a/web/server.go
+++ b/web/server.go
@@ -108,6 +108,8 @@ func Server() {
 	r.Use(middleware.RealIP)
 	r.Use(middleware.Logger)
 	r.Use(middleware.Recoverer)
+	r.Use(middleware.RedirectSlashes)
+	r.Use(middleware.StripSlashes)
 	r.Mount("/debug", middleware.Profiler())
 
 	r.Get("/", slaHTML)


web/server.go
```diff
@@ -108,8 +108,7 @@ func Server() {
 	r.Use(middleware.RealIP)
 	r.Use(middleware.Logger)
 	r.Use(middleware.Recoverer)
-	r.Use(middleware.RedirectSlashes)
-	r.Use(middleware.StripSlashes)
+	r.Mount("/debug", middleware.Profiler())
```
proditis, line 111: Hi @jordy1024, just a quick question, why the Mount "/debug" removes RedirectSlashes and StripSlashes (excuse my ignorance i'm fairly new with Go)

proditis: Never mind, I just saw your issue :+1:

jordy1024: > Hi @jordy1024, just a quick question, why the Mount "/debug" removes RedirectSlashes and StripSlashes (excuse my ignorance i'm fairly new with Go)

some issues about  `middleware.Profiler()`: https://github.com/go-chi/chi/issues/262

Maybe it's not the best solution. @proditis

haoel: Removing them is fine. And could you please test the following URls?

htttp://localhost:8181/sla
htttp://localhost:8181/sla/

htttp://localhost:8181/api/v1/sla
htttp://localhost:8181/api/v1/sla/

proditis: > Maybe it's not the best solution. @proditis

I cant think of anything wrong with it. As long as our SLA endpoints work I think this is an ok change.

jordy1024: > Removing them is fine. And could you please test the following URLs?
> 
> http://localhost:8181/sla http://localhost:8181/sla/
> 
> http://localhost:8181/api/v1/sla http://localhost:8181/api/v1/sla/
 
SO sorry, my bad.I'll think of a better solution.

haoel: Actually, it is really tough problem.  ;-)

1) the URL with a trailing slash or not.
2) if we put the easeprobe behind nginx.  `proxy_pass: http://easeprobe:8181` or  `proxy_pass: http://easeprobe:8181/` could introduce a double slash problem.

proditis:
ping! :smile: 

Hi @jordy1024, I was wondering what is the status of this PR? Its been stall for some time now, anything I can do to help?

PR #108, opened 2022-05-24
Connect probers and notifiers by Channel

haoel:
This PR introduces the channel to connect the probers and notifiers.  

For more information, please take a look the discussion  #82 

### Example:

if we have the following configuration

```yaml
http: 
   - name: probe A
     channels : [ Chann_A, Chann_B]
shell:
   - name: probe B
     channels: [ Chann_C ]
notify:
   - discord: Discord
     channels: [Chann_A, Chann_C]
   - email: Gmail
     channels: [Chann_B]
```
Then, we will have the following diagram

```
┌───────┐          ┌─────────┐
│Probe B├─────────►│ Chann_C ├─────┐
└───────┘          └─────────┘     │
                                   │
                                   │
                   ┌─────────┐     │    ┌─────────┐
            ┌─────►│ Chann_A ├─────▼────► Discord │
            │      └─────────┘          └─────────┘
┌───────┐   │
│Probe A├───┤
└───────┘   │
            │      ┌─────────┐          ┌─────────┐
            └─────►│ Chann_B ├──────────►  Gmail  │
                   └─────────┘          └─────────┘
```

### Test

I've done the following test:

- [x] Case 1: backward compatibility  - NO channels defined for all probers and notifiers.
- [x] Case 2 : 1 prober -> 1 channel -> 1 notifier. 
- [x] Case 3: 5 probers and 3 notifiers, define 3 channels.
- [x] Case 4: the channel hasn't probers.  - everything's fine, but a warning message in the log
- [x] Case 5: the channel hasn't notifiers - everything's fine, but a warning message in the log. 
- [x] Case 6: the probers and notifiers without channels that can be connected together.


zhao-kun:
What's the problem the PR will solve?

If we want to let users can config which notification can be applied to the probe, I think we shouldn't expose the channel concept to end-user, because it's technical and not intuitive. Can we give the notification a name and let it be referred in the probe?

haoel:
> What's the problem the PR will solve?
> 
> If we want to let users can config which notification can be applied to the prober, I don't think we should expose the `channel` concept to the end-user, because it's technical and not intuitive. Can we give the notification a name and let it be referred in one or more probers?

The channel is not the Go concept, it's the channel literally.  

To design like this, we have several considerations.

1) The requirement needs to group the notifiers, and probers just send the result to that group. 
2) if we configure the reference of notifiers in the probe, this would couple the notify to probe. it works, in some cases, it could bring complexity. for example, if we want to add or delete a notifier, we need to change many places. 

Theoretically, this is an N x M relationship model, introducing a referring object would be the best practice. 

You can take a look the discussion #82

global/global.go
```diff
@@ -59,6 +59,8 @@ const (
 	DefaultProbeInterval = time.Second * 60
 	// DefaultTimeOut is 30 seconds
 	DefaultTimeOut = time.Second * 30
+	// DefaultChannelName  is the default wide channel name
+	DefaultChannelName = "__EaseProbe_Channel__"
```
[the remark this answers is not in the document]
haoel: this would be considered the default channel, it would be grouped with all of the probers and notifiers who have not defined the channel.

zhao-kun:
> > What's the problem the PR will solve?
> > If we want to let users can config which notification can be applied to the prober, I don't think we should expose the `channel` concept to the end-user, because it's technical and not intuitive. Can we give the notification a name and let it be referred in one or more probers?
> 
> The channel is not the Go concept, it's the channel literally.
> 
> To design like this, we have several considerations.
> 
> 1. The requirement needs to group the notifiers, and probers just send the result to that group.
> 2. if we configure the reference of notifiers in the probe, this would couple the notify to probe. it works, in some cases, it could bring complexity. for example, if we want to add or delete a notifier, we need to change many places.
> 
> Theoretically, this is an N x M relationship model, introducing a referring object would be the best practice.
> 
> You can take a look the discussion #82 to have more information.

OK, Got it.  It's better to relate the issue with the PR in its description. So I can easy to know the background of the PR.

BTW, Why do we choose the `labels` and `selector` mechanism, the mechanism was practiced and applied in the K8s, and cloud systems widely.

channel/channel.go
```diff
@@ -0,0 +1,166 @@
...
+func SetDryNotify(dry bool) {
+	dryNotify = dry
+}
+
+// WatchEvent watches the notification event
+// Go through all of notification to notify the result.
+func (c *Channel) WatchEvent() {
+
+	// check if the channel is watching
+	if atomic.LoadInt32(&(c.isWatch)) != 0 {
```
localvar, line 126: Seems this line should use `atomic.CompareAndSwapInt32`.

haoel: I think you find a concurrent issue here. let me think about how to fix it.

jordy1024 (commented):
I mainly added some suggestions for improving code readability,for reference only.

jordy1024:
> > What's the problem the PR will solve?
> > If we want to let users can config which notification can be applied to the prober, I don't think we should expose the `channel` concept to the end-user, because it's technical and not intuitive. Can we give the notification a name and let it be referred in one or more probers?
> 
> The channel is not the Go concept, it's the channel literally.
> 
> To design like this, we have several considerations.
> 
> 1. The requirement needs to group the notifiers, and probers just send the result to that group.
> 2. if we configure the reference of notifiers in the probe, this would couple the notify to probe. it works, in some cases, it could bring complexity. for example, if we want to add or delete a notifier, we need to change many places.
> 
> Theoretically, this is an N x M relationship model, introducing a referring object would be the best practice.
> 
> You can take a look the discussion #82 to have more information.

Sounds like a very flexible, extensible design. 👍🏻

cmd/easeprobe/prober.go
```diff
@@ -0,0 +1,86 @@
...
+				if ch := channel.GetChannel(cName); ch != nil {
+					ch.Send(res)
+				}
+			}
+
+			select {
+			case <-done:
+				log.Infof("%s / %s - Received the done signal, exiting...", p.Kind(), p.Name())
+				return
+			case <-time.After(p.Interval()):
```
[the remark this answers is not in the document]
haoel: Using the `time.NewTimer()` =>  `defer timer.Stop()` and `timer.Reset()` pattern is much better.  I will refectory the code.

haoel:
> BTW, Why don't we choose the `labels` and `selector` mechanism, the mechanism was practiced and applied in the K8s, and cloud systems widely.

It's easy to do the naming work 
- change the notify's `channels` to `labels`
- change the probe's `channels` to `selector`

But I think the `channels` make more sense just like the TV Channel, as the notify just like boardcast. 

@proditis & @nullsimon, what do you think?

nullsimon:
> > BTW, Why don't we choose the `labels` and `selector` mechanism, the mechanism was practiced and applied in the K8s, and cloud systems widely.
> 
> 
> 
> It's easy to do the naming work 
> 
> - change the notify's `channels` to `labels`
> 
> - change the probe's `channels` to `selector`
> 
> 
> 
> But I think the `channels` make more sense as the notification just like the TV Channel.
> 
> 
> 
> @proditis & @nullsimon, what do you think? 


Lables and selector are great and commonly used,it's a kv store extension,and channels is just a one feature of them. If we have more new features,it is extensive.

channel/channel.go
```diff
@@ -0,0 +1,165 @@
...
+	"sync/atomic"
+)
+
+const kind = "channel"
+
+// Channel implements a config for Channel
+type Channel struct {
+	Name      string                    `yaml:"name"`      // unique name
+	Probers   map[string]*probe.Prober  `yaml:"probers"`   // probers
+	Notifiers map[string]*notify.Notify `yaml:"notifiers"` // notifiers
```
localvar, line 33:
```suggestion
	Probers   map[string]probe.Prober  `yaml:"probers"`   // probers
	Notifiers map[string]notify.Notify `yaml:"notifiers"` // notifiers
```

as both `probe.Prober` and `notify.Notify` are interfaces.

haoel:
I just realize using `labels` and `selector` is not just naming works. it's different. 

- For `channel`, both probe and notify can define that, we just create the channel arbitrarily. So, the code is simpler.
- For the `labels` and `selector`, only the `labels` can create the labels, and the `selector` won't and needs to check whether the label exists or not.


cc @zhao-kun @nullsimon @proditis

zhao-kun:
> > BTW, Why don't we choose the `labels` and `selector` mechanism, the mechanism was practiced and applied in the K8s, and cloud systems widely.
> 
> It's easy to do the naming work
> 
> * change the notify's `channels` to `labels`
> * change the probe's `channels` to `selector`
> 
> But I think the `channels` make more sense as the notification just like the TV Channel.
> 
> @proditis & @nullsimon, what do you think?

IMHO, the term channel represents different ways of notification, such as we could treat email as a channel, SMS as another channel. The label helps us group different channels flexibly. So these terms would be much more intuitive to users and easy to use if we adopt them.

nullsimon:
> I just realize using `labels` and `selector` is not just naming works. it's different. 
> 
> 
> 
> - For `channel`, both probe and notify can define that, we just create the channel arbitrarily. So, the code is simpler.
> 
> - For the `labels` and `selector`, only the `labels` can create the labels, and the `selector` won't and needs to check whether the label exists or not.
> 
> 
> 
> 
> 
> cc @zhao-kun @nullsimon @proditis 
> 
> 

Yeah,channel is just one dimension,labels and selector become two dimensions. For channel to probers and notifies is that they have a common base metadata,while another become a pub-sub thing. The More extensive the more complex.

proditis:
> * change the notify's `channels` to `labels`
> * change the probe's `channels` to `selector`
> 
> But I think the `channels` make more sense as the notification just like the TV Channel.

I like the common naming `channels` more also.  i'm not seen familiar with k8d so i dont know the context the `labels` and `selector` keywords are used there. 

I dont have strong arguments for or against the change, though, so i leave the final decision to you guys.

haoel:
First of all, both of them have the same flexibility.

Secondly, I believe different people like different terms, the intuition is subjective.

Channel is business-specific, and Label/Selector is more general. If we want to make business sense, we should use the Channel, otherwise using Label/Selector.

So, we must make the decision here. As the owner and PR author,  I will make the final decision - Using the Channel for this version ;-)  as this is based on the original discussion, and it's easier for coding work.  

Meanwhile, if we need the Label/Selector in the future, we still are open to doing that ...

README.md
```diff
@@ -233,7 +234,56 @@ Check the  [Notification Configuration](#37-notification-configuration) to see h
...
+
+The Channel is used for connecting the Probers and the Notifiers. It can be configured for every Prober and Notifier.
+
+This feature could help you group the Probers and Notifiers into a logical group.
+
+> **Note**:
+> 
+> 1) If you don't define the Channel, the default channel will be used for these probers and notifiers. The default channel name is `__EaseProbe_Channel__` 
+> 
+> 2) Versions of EaseProbe prior to  v1.5.0, do not support the `channel` feature
```
proditis, line 247: I think we can remove the note from this spot and add it to the release notes of the next version.

README.md
```diff
@@ -233,7 +234,56 @@ Check the  [Notification Configuration](#37-notification-configuration) to see h
...
+
+```YAML
+
+For example:
+
+```YAML
+http: 
+   - name: probe A
+     channels : [ Dev_Channel, Manager_Channel ]
```
proditis, line 256: Do the channel names need to follow any particular rules? 
Such as no space, no symbols etc?

README.md
```diff
@@ -233,7 +234,56 @@ Check the  [Notification Configuration](#37-notification-configuration) to see h
...
+shell:
+   - name: probe B
+     channels: [ Ops_Channel]
```
proditis, line 259:
```suggestion
     channels: [ Ops_Channel ]
```

README.md
```diff
@@ -233,7 +234,56 @@ Check the  [Notification Configuration](#37-notification-configuration) to see h
...
+notify:
+   - discord: Discord
+     channels: [Dev_Channel, Ops_Channel]
```
proditis, line 262:
```suggestion
     channels: [ Dev_Channel, Ops_Channel ]
```

README.md
```diff
@@ -233,7 +234,56 @@ Check the  [Notification Configuration](#37-notification-configuration) to see h
...
+   - email: Gmail
+     channels: [Mgmt_Channel]
```
proditis, line 264:
```suggestion
     channels: [ Mgmt_Channel ]
```

proditis: a few spaces just for consistency

proditis (approved):
LGTM

tg123:
IHMO, labels and selector, used in k8s, might be better approach for the scenario.

README.md
haoel: looks like the limitation is just only for yaml file.  If the chars can be accepted by yaml, then we can support that.

localvar (approved):

README.md
haoel: This is the major important feature, I think it's better to mention it in the user manual.  like other software.

PR #110, opened 2022-05-26
introduce tls probe

tg123:
TLS ping to remote endpoint, can probe for revoked or expired certificates

### DIFF
--- a/probe/tls/tls.go
+++ b/probe/tls/tls.go
@@ -14,6 +14,7 @@
  * See the License for the specific language governing permissions and
  * limitations under the License.
  */
+
 package tls
 
 import (
--- a/probe/tls/tls.go
+++ b/probe/tls/tls.go
@@ -1,0 +1,120 @@
+/*
+ * Copyright (c) 2022, MegaEase
+ * All rights reserved.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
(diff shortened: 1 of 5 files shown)


probe/tls/tls.go
```diff
@@ -0,0 +1,120 @@
...
+)
+
+type TLS struct {
+	base.DefaultOptions `yaml:",inline"`
+	Host                string `yaml:"host"`
+	InsecureSkipVerify  bool   `yaml:"insecure_skip_verify"`
+
+	RootCAPemPath string `yaml:"root_ca_pem_path"`
+	RootCaPem     string `yaml:"root_ca_pem"`
+	rootCAs       *x509.CertPool
```
haoel, line 42: There is a `TLS` struct and `Config()` method in `global/global.go`, do you think we can reuse that?

tg123: never know that
but the global one introduces extra prop `cert` any `key` which is not only useless but also creating confusion
client cert is not required for tls test

haoel: the global TLS is originally designed for mTLS, so do you think we can adapt this tls probe for both general TLS and mTLS?

haoel: Just realized this PR only detects the expiration of certificate, pls ignore the mTLS comments.

And pls add the metric for Prometheus, thanks!

haoel (approved):
The code looks fine to me! Ship it!

PR #111, opened 2022-05-26
notify to support ctx

tg123:
allow notify manager to control timeout of notify impl

follow ups:
 * all notify impls should honor ctx
 * caller should replace context.TODO with correct ctx (maybe a timeout)

### DIFF
--- a/cmd/easeprobe/main.go
+++ b/cmd/easeprobe/main.go
@@ -18,6 +18,7 @@
 package main
 
 import (
+	"context"
 	"flag"
 	"fmt"
 	"os"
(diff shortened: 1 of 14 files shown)


haoel:
Thanks for this PR. But I think we can do the timeout works for all of the Notify in the base notifier, it's totally transparent for all notifiers, just like the `global.DoRetry()`

```go
func DoTimeout(timeout time.Duration, func sendFn() error)
	ctx, cancel := context.WithTimeout(context.Background(), timeout)
	defer cancel()
	done := make(chan error, 1)
	go func() {
		done <- sendFn()
	}()

	select {
	case err := <-done:
		return err
	case <-ctx.Done():
		return ctx.Err()
	}
}
```

tg123:
goroutine should stop when ctx being canceled
otherwise, would cause leak

haoel:
since we don't have the plan to use the context to manage all of the notifications and probes, I would close this PR at this moment.

PR #136, opened 2022-06-09
Add `env` and `clean_env`  for shell notification and probe

haoel:
if set the `clean_env:true`, the EaseProbe won't pass the OS environment variables to the shell notification and probe. the default value is false


proditis (approved):
Tested and reviewed 
LGTM

PR #138, opened 2022-06-09
Memcache client

proditis:
This PR adds support for Memcache client probe
* [x] Add support for `data:` on all client configs
* [x] Check for key(s) when conf `data:` option exists
* [x] Check for value (optional) when conf option exists eg `key: val`
* [x] If no key exists issue a ping command instead

```yaml
client:
  - name: Memcache
    driver: "memcache"
    host: "localhost:11211" # host to query
    data:
      "sysconfig:event_active": "1" # Check the namespaced key `sysconfig:event_active` has value `1` and fail if not
      mykey: "value" # Check the key `mykey` has value `value` and fail if not
      mykeywithnovalcheck:  # Just check the existence of key
```

@haoel i am currently stuck with this. I will need help with the following
1. I cant figure out how to make it populate the Key/Value pairs from the config my m.Key and m.Value are always nil
2. I cant figure out how to utilize ctx with the Memcache Get() and New()

### DIFF
--- a/probe/client/memcache/memcache.go
+++ b/probe/client/memcache/memcache.go
@@ -97,12 +97,12 @@ func (m *Memcache) validateKeyValues(items map[string]*MemcacheClient.Item) (boo
 	// iterate the keys and confirm their values match
 	for _, item := range items {
 		log.Debugf("Got key: %s with value: %s", item.Key, string(item.Value))
-		if string(m.Data[item.Key]) != "" {
-			if string(item.Value) != m.Data[item.Key] {
-				return false, fmt.Sprintf("Memcache value for key %s returned %s, expected %s", item.Key, string(item.Value), string(m.Data[item.Key]))
-			}
-		} else {
+		if strings.TrimSpace(m.Data[item.Key]) == "" {
 			log.Debugf("Skipping value check for item %s", item.Key)
+			continue
+		}
+		if string(item.Value) != m.Data[item.Key] {
+			return false, fmt.Sprintf("Memcache value for key %s returned %s, expected %s", item.Key, string(item.Value), string(m.Data[item.Key]))
 		}
 	}
 	return true, "Memcache key values match"
--- a/probe/client/memcache/memcache.go
+++ b/probe/client/memcache/memcache.go
@@ -1,0 +1,70 @@
+/*
+ * Copyright (c) 2022, MegaEase
+ * All rights reserved.
+ *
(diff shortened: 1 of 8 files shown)


haoel:
@proditis ,  I'm sorry I misled you yesterday.

The `client` probe's configuration cannot be extended by itself, the configuration has to be done by `client.Options` in `client/conf/conf.go` file.

For your case, I think we can add a field to `client.Options` - 
```go
type Options struct {
    ... ...
    Data map[string]string `yaml:"data"`
   ... ...
}
```

Then, we can have the following configurations:

```yaml
  - name: Memcache
    driver: memcache
    host: localhost: 11211
    data:
      key1: value1
      key2: value2
      key3: value3
```

we can read the key/value data like this:  `Memcache.Data["key1"]`, `Memcache.Data["key2"]`, `Memcache.Data["key3"]`



for the timeout setting, after reading the [source code](https://github.com/bradfitz/gomemcache/blob/fb4bf637b56d66a1925c1bb0780b27dd714ec380/memcache/memcache.go#L137), I think you can simply just set our timeout to its client timeout

```go
mc := memcache.New()
mc.Timeout = m.Timeout()
```

currently, it doesn't support new a client with a context. there are a couple of PRs related to Context that hasn't been accepted yet, and the TLS as well, and I even cannot find any other alternative Memcache client, sad!

probe/client/memcache/memcache.go
```diff
@@ -0,0 +1,109 @@
...
+	// iterate the keys and confirm their values match
+	for _, item := range items {
+		log.Debugf("Got key: %s with value: %s", item.Key, string(item.Value))
+		if string(m.Data[item.Key]) != "" {
+			if string(item.Value) != m.Data[item.Key] {
+				return false, fmt.Sprintf("Memcache value for key %s returned %s, expected %s", item.Key, string(item.Value), string(m.Data[item.Key]))
+			}
+		} else {
+			log.Debugf("Skipping value check for item %s", item.Key)
+		}
```
haoel, line 106:
```suggestion
		if strings.TrimSpace(m.Data[item.Key]) == "" {
			log.Debugf("Skipping value check for item %s", item.Key)
			continue
		}
		if string(item.Value) != m.Data[item.Key] {
			return false, fmt.Sprintf("Memcache value for key %s returned %s, expected %s", item.Key, string(item.Value), string(m.Data[item.Key]))
		}
```

haoel (approved):
LGTM

PR #141, opened 2022-06-15
disable CGO to build static executable

hellojukay:


haoel (approved):
Thanks for this PR. Disable the CGO in the makefile is fine. 

P.S. the EaseProbe release binaries are built by goreleaser(refer to the [.goreleaser.yaml](https://github.com/megaease/easeprobe/blob/main/.goreleaser.yaml)), which is disable the CGO

PR #142, opened 2022-06-16
add go ver and git sha to ver

tg123:
before
`EaseProbe v1.6.0`
after
`EaseProbe v1.6.0 afb3f189c 2022-06-16T09:17:06Z go1.18.3`


proditis:
Hi @tg123 

I think this enables debug logging everywhere. check the attached screenshot
image

I think this is caused by the `log.SetLevel(log.DebugLevel)`, can we change this to only activate when the `-v` option is given? 

I think this is also good to include into the notifications footer of discord, when debug is enabled?
image

Other than these two comments looks good to me :+1:

tg123:
mb, I commit debug by accident
will add log level in separate pr

cmd/easeprobe/main.go
```diff
@@ -43,8 +44,30 @@ func getEnvOrDefault(key, defaultValue string) string {
...
-func main() {
+func showVersion() {
+
+	var v = global.Ver
+
+	bi, ok := debug.ReadBuildInfo()
+	if !ok {
+		v = fmt.Sprintf("%v %v", global.DefaultProg, v)
+		fmt.Println(v)
+	}
```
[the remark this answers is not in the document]
tg123: this is the old ver format without gitsha, otherwise -v returns nothing looks weird 

i do not know when will it be `false`, seems not being built in git repo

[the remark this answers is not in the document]
haoel: if the `debug.ReadBuildInfo()` return the `false`, I believe the `bi` is nil, if you don't return here, the following code would be panic with the nil pointer access error. right?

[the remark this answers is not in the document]
tg123: good catch
mb
this is copied from my another repo and was `rtn str`
return was replaced by printf by mistake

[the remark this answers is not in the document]
haoel: are you dunk during the coding? 😂

haoel:
I'm thinking you probably can store that information in `EaseProbe` struct in `global/easeprobe.go`, and add them into the `FooterString()` , then the notification would see such information just like @proditis mentioned above.

tg123:
is it too long?
image

global/global.go
```diff
@@ -223,3 +224,24 @@ func CommandLine(cmd string, args []string) string {
 	}
 	return result
 }
+
+func VersionString() string {
```
haoel, line 228: Lint: exported function VersionString should have comment or be unexported

global/easeprobe.go
```diff
@@ -43,7 +43,7 @@ func InitEaseProbe(name, icon string) {
 	easeProbe = &EaseProbe{
 		Name:    name,
 		IconURL: icon,
-		Version: Ver,
+		Version: VersionString(),
```
haoel, line 46: I think it's better to have individual fields here. 

AppVersion - v1.6.0
GoVersion - v1.18.2
VCSReversion - afcdefg
VCSTime - 2022-....

Then, it can be easy to pick up the interests.  For example, we can get rid of the VCSTime in FootString...

tg123: that way, it should call `runtime/debug` directly

I prefer not to add sha1 to footer

tg123: either of below are too ugly/confused to me
EaseProbe v1.6.0 24a5f9109 2022-06-17T04:43:23Z go1.18.3 @ OMEN at 2022-06-17 08:57:39 UTC (timestamp ambiguity)
EaseProbe v1.6.0 24a5f9109 go1.18.3 @ OMEN at 2022-06-17 08:57:39 UTC (version ambiguity)
EaseProbe v1.6.0 24a5f9109 @ OMEN at 2022-06-17 08:57:39 UTC (why a sha1 here)

haoel: Yes, it seems a bit messy. I think it's OK we keep the current notification footer with no change. What do you think?  @proditis

haoel: Cool, let approve this PR

haoel (approved):

PR #143, opened 2022-06-16
support log level

tg123:
```
  -l string
        log level, one of: trace, debug, info, warn, error, fatal, panic (default "info")
```


haoel:
We have a log configuration in settings already, if we need to support the log level in the command line, we have to carefully define the behavior if they are conflicts. for example, the command line gives debug, but the config.yml  gives info, which one would be the final decision?

Furthermore, we have a Log structure in conf/log.go which has comprehensive log works, I suggest merging this command line flag to the config structure - `conf.Settings.Log`, and the `c.InitAllLogs()` in the main function would help to do everything to initialize the log.

tg123:
never know there is a setting

how does this work
https://github.com/megaease/easeprobe/pull/142/commits/aee3e5e8757a8072c932d21a2bd5afe88eb65ac9

likely should be override by default right?

haoel:
may I know why your want the debug level log when the project start?

PR #145, opened 2022-06-20
Add an option to disable the HTTPS security checks

haoel:


README.md
```diff
@@ -493,6 +493,8 @@ http:
     ca: /path/to/file.ca
     cert: /path/to/file.crt
     key: /path/to/file.key
+    # TLS
+    insecure: true # skip the security check, use for self-signed certs. default: false
```
proditis, line 497:
```suggestion
    insecure: true # skip any security checks, useful for self-signed and expired certs. default: false
```

global/global.go
```diff
@@ -139,13 +145,25 @@ func (t *TLS) Config() (*tls.Config, error) {
 	caCertPool := x509.NewCertPool()
 	caCertPool.AppendCertsFromPEM(cert)
 
+	// only have CA file, go TLS
+	if len(t.Cert) <= 0 || len(t.Key) <= 0 {
+		log.Debug("[TLS] Only have CA file, go TLS")
+		return &tls.Config{
+			RootCAs:            caCertPool,
```
zhao-kun, line 152: x509.NewCertPool creates a new empty cert pool.  
x509.SystemCertPool returns a copy of the system cert pool.

I think NewCertPool will override the default system CA configurations, which would make the legitimate certs signed by public authorities verify failed. Is it your expectation?

haoel: sorry, I merge this pr so quickly.

i think the better way should be 

```diff
@@ -142,7 +142,10 @@ func (t *TLS) Config() (*tls.Config, error) {
        if err != nil {
                return nil, err
        }
-       caCertPool := x509.NewCertPool()
+       caCertPool, _ := x509.SystemCertPool()
+       if caCertPool == nil {
+               caCertPool = x509.NewCertPool()
+       }
        caCertPool.AppendCertsFromPEM(cert)
 
        // only have CA file, go TLS
```

I will open another PR to address this.

tg123: if cert then create new pool
else dothing and goruntime will use system pool

should not call add to system pool

PR #146, opened 2022-06-21
bug-fix: command line option dry notification not working

haoel:
when I refactor the code to support the channel, I missed this function.


proditis (approved):
tested and works as expected
LGTM

PR #147, opened 2022-06-23
feat: support multiple disks usage check for host probe

haoel:
the original host probe only supports root disk usage check.

the PR allows user to configure multiple disks. if not configure the "disks", the default behavior just check the root fs.

```yaml
host:
  servers:
    - name: server
      host: user@1.1.1.1
      ....
      disks:
        - /
        - /data
 ```

### DIFF
--- a/probe/host/host.go
+++ b/probe/host/host.go
@@ -78,7 +78,7 @@ func (s *Server) Config(gConf global.ProbeSettings) error {
 	// 4. retrieve the cpu core:		`grep -c ^processor /proc/cpuinfo;`
 	// 5. retrieve the cpu usage:	`top -b -n 1 | grep Cpu | awk -F ":" '{print $2}'`
 	//    output example: 1.6 us,  0.0 sy,  0.0 ni, 98.4 id,  0.0 wa,  0.0 hi,  0.0 si,  0.0 st
-	// 6. retrieve the disk usage	`df -h / 2>/dev/null | awk '$NF=="/"{printf "%d %d %s\n", $3,$2,$5,$6}'`
+	// 6. retrieve the disk usage	`df -h / 2>/dev/null | awk '(NR>1){printf "%d %d %s %s\n", $3,$2,$5,$6}'`
 	//    output: used(GB) total(GB) usage(%) disk, example: 40 970 5% /
 
 	s.Command = `hostname;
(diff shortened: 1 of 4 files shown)

### CODE REVIEW

File: probe/host/host.go [hunk 1]
```diff
@@ -84,8 +85,14 @@ func (s *Server) Config(gConf global.ProbeSettings) error {
 	awk -F= '/^NAME/{print $2}' /etc/os-release | tr -d '\"';
```
zhao-kun, line 85: Need to update the comment on item 6 because you add a new column to output

probe/host/host.go
```diff
@@ -145,12 +156,14 @@ func (s *Server) CheckThreshold(info Info) (bool, string) {
...
-		status = false
-		if message != "" {
-			message += " | "
+	for _, disk := range info.Disks {
+		if s.Threshold.Disk > 0 && s.Threshold.Disk <= disk.Usage/100 {
+			status = false
+			if message != "" {
+				message += " | "
+			}
+			message += fmt.Sprintf("Disk Full! - [%s]", disk.Tag)
```
proditis, line 165:
```suggestion
			message += fmt.Sprintf("Disk usage exceeds threshold! - [%s]", disk.Tag)
```

probe/host/host.go
```diff
@@ -77,15 +78,21 @@ func (s *Server) Config(gConf global.ProbeSettings) error {
...
 	grep -c ^processor /proc/cpuinfo;
-	top -b -n 1 | grep Cpu | awk -F ":" '{print $2}';
-	df -h / 2>/dev/null | awk '$NF=="/"{printf "%d %d %s\n", $3,$2,$5}'`
+	top -b -n 1 | grep Cpu | awk -F ":" '{print $2}';` + "\n"
+
+	if len(s.Disks) == 0 {
+		s.Disks = []string{"/"}
+	}
+	for _, disk := range s.Disks {
+		s.Command += "\t" + `df -h ` + disk + ` 2>/dev/null | awk '$NF=="` + disk + `"{printf "%d %d %s %s\n", $3,$2,$5,$6}';` + "\n"
```
proditis, line 94: We can implode the disks into a single command such as `dh -h / /data` and adapt the awk accordingly
```shell
df / /media 2>/dev/null | awk '(NR>1){printf "%d %d %s %s\n", $3,$2,$5,$6}'
301876116 459923528 70% /
666395536 960385056 74% /media
```

probe/host/host.go
proditis: but the threshold does not necessarily mean that the disk is full :(
I believe that since we define our own thresholds these messages need to be adapted to state that. 
As an example my threshold is at 70% so when i pass the threshold my disks are not Full.

haoel: I mean we should give the same pattern for all of the resource usages.

And considering the thresholds we would use to determine the health(up/down) of the host, so, it's better to give the related message. 

I agree the "Full" might be too serious, maybe we can say: Disk Space Low!

probe/host/host.go
haoel: Good, let me change the code.

probe/host/host.go
proditis: > I mean we should give the same pattern for all of the resource usages.
Yes i agree with that, i'd like the messages to be consistent also. 

Lets ignore this particular suggestion so as to proceed with the main PR which is to add support for disks. We can address these messages in another PR to make sure the messages are consistent and more appropriate for each case?

haoel: no problem

proditis (approved):
LGTM

PR #148, opened 2022-06-23
Using DurationStr() format Discord Uptime and DownTime

haoel:
Currently, the SLA report in Discord reports the Uptime and Downtime in the following format:

Up: `785h39m0s`   <- 785 hours is too large

It should be: 

Up:  `32d17h39m0s`  <- 32 days and 17 hours would be better.

The second format is used by all notifications.


proditis (approved):
image
LGTM

zhao-kun (approved):

PR #152, opened 2022-07-01
Add some filter options for SLA HTML Page

haoel:
For the HTML report, add the following options:

  - `status=up`: only shows the probers which status is `up`. the value can be `up` or `down`.
  - `gte=50`: only shows the probers the SLA percentage is >= `50%`.
  - `lte=90`: only shows the probers the SLA percentage s  <= `90%`.

PR #158, opened 2022-07-03
feat: Add more query filter options for SLA HTML and JSON API

haoel:
this PR is the enhancement for PR #152 

we make the filter more general and add more query options for both SLA HTML and JSON

the following options are added:
 -  `pg` & `sz`: page number and page size, , default page size is `100` (ex, `?pg=2&sz=10` shows the second page with 10 probers)
  - `name`: filter the probers with the name (ex, `?name=probe1` list the probers which name containing `probe1`)
  - `kind`: filter the probers with the kind (ex, `?kind=http` list the probers with kind `http`)
  - `ep`: filter the probers with the endpoint (ex, `?ep=example.com` list the probers which endpoint containing  `example.com`)
  - `msg`: filter the probers with the message (ex, `?msg=example` list the probers which message containing `example`)

example:

query string `http://localhost:8181/?status=up&gte=60&msg=Success&sz=3&pg=2`

image

### DIFF
--- a/README.md
+++ b/README.md
@@ -257,6 +257,7 @@ Check the  [Notification Configuration](#38-notification-configuration) to see h
 
 You can use the following URL query options for both HTML and JSON:
   - `refresh`: (_HTML only_) refresh the page every given seconds (ex, `?refresh=30s` refreshes the page every 30 seconds)
+  - `pg` & `sz`: page number and page size (ex, `?pg=2&sz=10` shows the second page with 10 probers), default page size is `100`
   - `name`: filter the probers that contain the value of name (ex, `?name=probe1` list the probers which name containing `probe1`)
   - `kind`: filter the probers with the kind (ex, `?kind=http` list the probers with kind `http`)
   - `ep`: filter the probers with the endpoint (ex, `?ep=example.com` list the probers which endpoint containing  `example.com`)
--- a/README.md
+++ b/README.md
@@ -257,7 +257,7 @@ Check the  [Notification Configuration](#38-notification-configuration) to see h
 
 You can use the following URL query options for both HTML and JSON:
   - `refresh`: (_HTML only_) refresh the page every given seconds (ex, `?refresh=30s` refreshes the page every 30 seconds)
-  - `name`: filter the probers with the name (ex, `?name=probe1` list the probers which name containing `probe1`)
+  - `name`: filter the probers that contain the value of name (ex, `?name=probe1` list the probers which name containing `probe1`)
   - `kind`: filter the probers with the kind (ex, `?kind=http` list the probers with kind `http`)
   - `ep`: filter the probers with the endpoint (ex, `?ep=example.com` list the probers which endpoint containing  `example.com`)
   - `msg`: filter the probers with the message (ex, `?msg=example` list the probers which message containing `example`)
--- a/README.md
+++ b/README.md
@@ -252,15 +252,16 @@ Check the  [Notification Configuration](#38-notification-configuration) to see h
 
   The EaseProbe would listen on the `0.0.0.0:8181` port by default. And you can access the Live SLA report by the following URL:
 
-  - HTML: `http://localhost:8181/`
+  - HTML: `http://localhost:8181/` ( the query option `?refresh=30s` refreshes the page every 30 seconds))
   - JSON: `http://localhost:8181/api/v1/sla`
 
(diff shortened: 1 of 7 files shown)


README.md
```diff
@@ -252,15 +252,17 @@ Check the  [Notification Configuration](#38-notification-configuration) to see h
...
   - JSON: `http://localhost:8181/api/v1/sla`
 
-For the HTML report, you can use the following URL query options:
-
-  - `refresh`: report refresh rate (ex. `?refresh=30s` refreshes the page every 30 seconds)
-  - `status`: only shows the probers with specific status, accepted values `up` or `down` (ex. `?status=up` list only probers with status `up`).
-  - `gte`: only shows the probers with SLA greater than or equal to the given percentage (ex. `?gte=50` filter only hosts with SLA percentage `>= 50%`)
-  - `lte`: only shows the probers with SLA less than or equal to the given percentage (ex. `?lte=90` filter only hosts with SLA percentage `<= 90%` )
+You can use the following URL query options for both HTML and JSON:
+  - `name`: filter the probers with the name (ex, `?name=probe1` list the probers which name containing `probe1`)
```
proditis, line 259:
```suggestion
  - `name`: filter the probers that contain the value of name (ex, `?name=probe1` list the probers which name containing `probe1`)
```

README.md
```diff
@@ -252,15 +252,17 @@ Check the  [Notification Configuration](#38-notification-configuration) to see h
 
   The EaseProbe would listen on the `0.0.0.0:8181` port by default. And you can access the Live SLA report by the following URL:
 
-  - HTML: `http://localhost:8181/`
+  - HTML: `http://localhost:8181/` ( the query option `?refresh=30s` refreshes the page every 30 seconds))
```
proditis, line 255: ? why did we move the refresh back here?

proditis (commented):
* `kind=memcache` didnt work. http and tcp worked

README.md
haoel: this is only for the HTML page, not for the JSON API. so...

haoel: You can make any change as you like @proditis , I just make the change quickly without thinking better idea...

proditis, line 255:
```suggestion
  - HTML: `http://localhost:8181/`
```

README.md
proditis, line 259:
```suggestion
  - `refresh`: (_HTML only_) refresh the page every given seconds (ex, `?refresh=30s` refreshes the page every 30 seconds)
  - `name`: filter the probers with the name (ex, `?name=probe1` list the probers which name containing `probe1`)
```

README.md
proditis: how about these?

haoel:
> * `kind=memcache` didnt work. http and tcp worked
> 
> 

`kind=Mencache` ?

haoel:
> > * `kind=memcache` didnt work. http and tcp worked
> 
> `kind=Memcache` ?

the memcache kind is the `client`., and the memcache is stored in the probe tag... 

probably we need to add the tag option for query.

report/filter.go
```diff
@@ -0,0 +1,170 @@
...
+}
+
+// Filter filter the probers
+func (f *SLAFilter) Filter(probers []probe.Prober) []probe.Prober {
+	start, end := f.getIndics()
+	f.cnt = 0
+	result := make([]probe.Prober, 0)
+	for _, p := range probers {
+		// if the name is not empty then filter by name
+		if strings.TrimSpace(f.Name) != "" && !strings.Contains(p.Name(), f.Name) {
```
zhao-kun, line 138: Trimming of properties of the Filter in the loop is inefficient,  it should be better to trim these fields out of the loop.

web/server.go
```diff
@@ -63,53 +67,74 @@ func getFloat(f string, _default float64) float64 {
...
-	interval := getRefreshInterval(req.URL.Query().Get("refresh"))
-
-	filter := report.SLAFilter{}
+func getFilter(req *http.Request) (*report.SLAFilter, error) {
+	filter := &report.SLAFilter{}
 
-	if req.URL.Query().Get("status") != "" {
-		filter.Status = getStatus(req.URL.Query().Get("status"))
-	}
+	filter.Name = req.URL.Query().Get("name")
```
zhao-kun, line 91: It's better to trim the value here

zhao-kun (approved):

PR #160, opened 2022-07-03
Change the start up delay if we have less than 60 probers

proditis:
This PR changes the start up delay of the probers.

if probers are less than 60 then just use 1 second delay instead


cmd/easeprobe/probe.go
```diff
@@ -58,6 +58,9 @@ func configProbers(probers []probe.Prober) []probe.Prober {
 func runProbers(probers []probe.Prober, wg *sync.WaitGroup, done chan bool) {
 	// we need to run all probers in equally distributed time, not at the same time.
 	timeGap := global.DefaultProbeInterval / time.Duration(len(probers))
+	if len(probers) <= 60 {
+		timeGap = 1000000000
```
haoel, line 62:
```suggestion
		timeGap = time.Second
```

haoel (approved):

PR #170, opened 2022-07-16
bug-fix: Lark(Feishu) notifications don't work

xhkyyy:
This PR fix fails to send a notification to Lark(Feishu): `code: 9499, msg: bad request`.
This error is found in the Feishu document: https://open.feishu.cn/document/ukTMukTMukTM/ugjM14COyUjL4ITN


haoel:
Thanks for this PR.  So, the root cause is that the request body's JSON is invalid because of lacking the quotes, right?

xhkyyy:
> Thanks for this PR. So, the root cause is that the request body's JSON is invalid because of lacking the quotes, right?

@haoel yes, and the `StatusCode` check is also wrong, it causes always an error value to be returned instead of `nil`.

haoel (approved):

PR #171, opened 2022-07-17
feat: Support MySQL/Redis/MongoDB/PostgreSQL/Zookeeper Native Client  to check the data

haoel:
## Redis

```yaml
client:
  - name: Redis Native Client (local)
    driver: "redis"  # driver is redis
    host: "localhost:6379"  # server and port
    data:         # Optional
      key: val    # Check that `key` exists and its value is `val`
```
## MySQL

```yaml
client
   - name: MySQL Native Client (local)
      driver: "mysql"
      host: "localhost:3306"
      data: # Optional, check the specific column value in the table
        #  Usage: "database:table:column:primary_key:value" : "expected_value"
        #         transfer to : "SELECT column FROM database.table WHERE primary_key = value"
        #         the `value` for `primary_key` must be int
        "test:product:name:id:1" : "EaseProbe" # select name from test.product where id = 1
        "test:employee:age:id:2" : 45          # select age from test.employee where id = 2

```
##  MongoDB

```yaml
client:
  - name: MongoDB Native Client (local)
    driver: "mongo"
    host: "localhost:27017"
    data: # Optional, find the specific value in the table
      #  Usage: "database:collection" : "{JSON}"
      "test:employee" : '{"name":"Hao Chen"}' # find the employee with name "Hao Chen"
      "test:product" : '{"name":"EaseProbe"}' # find the product with name "EaseProbe"
```
## PostgreSQL
```yaml
client:
  - name: PostgreSQL Native Client (local)
    driver: "postgres"
    host: "localhost:5432"
    data: # Optional, check the specific column value in the table
      #  Usage: "database:table:column:primary_key:value" : "expected_value"
      #         transfer to : "SELECT column FROM table WHERE primary_key = value"
      #         the `value` for `primary_key` must be int
      "test:product:name:id:1" : "EaseProbe" # select name from product where id = 1
      "test:employee:age:id:2" : 45          # select age from employee where id = 2
```

## Zookeeper

```yaml
client:
  - name: Zookeeper Native Client (local)
    driver: "zookeeper"
    host: "localhost:2181"
    data: # Optional, check the specific value in the path
      "/path/to/key": "value" # Check that the value of the `/path/to/key` is "value"
```

### DIFF
--- a/probe/client/mongo/mongo.go
+++ b/probe/client/mongo/mongo.go
@@ -22,7 +22,6 @@ import (
 	"fmt"
 	"strings"
 
-	"github.com/megaease/easeprobe/global"
 	"github.com/megaease/easeprobe/probe/client/conf"
 	log "github.com/sirupsen/logrus"
 	"go.mongodb.org/mongo-driver/bson"
@@ -43,7 +42,7 @@ type Mongo struct {
 }
 
 // New create a Mongo client
-func New(opt conf.Options) Mongo {
+func New(opt conf.Options) (*Mongo, error) {
 	var conn string
 	if len(opt.Password) > 0 {
 		conn = fmt.Sprintf("mongodb://%s:%s@%s/?connectTimeoutMS=%d",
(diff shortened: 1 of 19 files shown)


probe/client/mongo/mongo.go
```diff
@@ -50,6 +52,8 @@ func New(opt conf.Options) Mongo {
 			opt.Host, opt.Timeout().Milliseconds())
 	}
 
+	log.Debugf("[%s / %s / %s] - Connection - %s", opt.ProbeKind, opt.ProbeName, opt.ProbeTag, conn)
```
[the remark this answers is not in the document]
haoel: I think this should be fine because the config.yaml also contains the username & password.

and this is the debug log, by default, it is disabled. and to enable that, we need to configure the config.yaml, and the config.yaml exposes everything.

suchen-sci (approved):

zhao-kun:
Question: Do we need to support probing all nodes of a Redis cluster with one configuration?

probe/client/mongo/mongo.go
```diff
@@ -60,27 +64,50 @@ func New(opt conf.Options) Mongo {
...
+		for k, v := range r.Data {
+			if _, _, err := getDBCollection(k); err != nil {
+				return err
+			}
+			var bdoc interface{}
+			if err := bson.UnmarshalExtJSON([]byte(v), true, &bdoc); err != nil {
+				return err
+			}
+		}
+	}
```
localvar, line 105: propose to remove `if` to make code simpler.
```suggestion
	for k, v := range r.Data {
		if _, _, err := getDBCollection(k); err != nil {
			return err
		}
		var bdoc interface{}
		if err := bson.UnmarshalExtJSON([]byte(v), true, &bdoc); err != nil {
			return err
		}
	}
```

probe/client/mysql/mysql.go
```diff
@@ -53,25 +55,43 @@ func New(opt conf.Options) MySQL {
...
 
 // Kind return the name of client
-func (r MySQL) Kind() string {
+func (r *MySQL) Kind() string {
 	return Kind
 }
 
+// checkData do the data checking
+func (r *MySQL) checkData() error {
+	if len(r.Data) > 0 {
```
localvar, line 83: ditto

proditis (commented):
LGTM not tested yet

haoel:
> Question: Do we need to support probing all nodes of a Redis cluster with one configuration?

At this moment, we don't have a plan to do this, we only can configure a cluster by individual configuration. 

To introduce the cluster concept is good for configuration, but we have to  separate the cluster into individual prober, and name every prober a unique name, and people might need a cluster-wide healthy status or SLA, this would bring complexity to the current design. 

So, I'd like to be a businessman here, I want to see how many benefits we can get from this requirement. ;-)

muicoder:
LGTM!

PR #173, opened 2022-07-20
change the getInt() and getFloat() to generic function

haoel:


localvar (approved):
I'm fine with the code changes, but I think the prior version is simpler and more explicit.

zhao-kun (approved):

PR #190, opened 2022-08-07
feat: support Dingtalk notify for sign secret

samanhappy:


haoel (approved):
Thanks for this contribution.

PR #192, opened 2022-08-10
feat: add the XPath/RegEx Data Extractor and Expression Evaluation for HTTP Probe

haoel:
try to support the following expression evaluation.

```yaml
http:
  - name: EaseProbe RSS Feed
    url: https://github.com/megaease/easeprobe/releases.atom
    eval:
      doc : XML # <-- document type, currently support HTML, XML, JSON, TEXT
      expression: "updated > '2022-07-01'"  # <-- the expression need to be evaluated.
      # define how the variable `updated` come from
      variables: 
         - name: updated
           type: time # support int/float/string/bool/time/duration 
           query: "//feed/updated" # XPath 
           time_format: "2006-01-02T15:04:05Z07:00"
```

or simple one 

```yaml
http:
  - name: EaseProbe RSS Feed
    url: https://github.com/megaease/easeprobe/releases.atom
    proxy: socks5://localhost:1085
    eval:
      doc : XML
      expression: "x_time('//feed/updated') > '2022-07-01'"
```
**Note**: `x_str()`, `x_int()`, `x_float()`, `x_time()`, `x_bool()`, `x_duration()` are the functions can extract the data by using XPath or Regex expression

Please focus on those files
- eval/*.go
- probe/http/http.go

> **Note**: 
>
> There are many files that have been changed because of the go 1.19 format.


zhao-kun (commented):
IMHO, If I implement the feature, I would prefer to splitting the Evaluator and Configuration into different objects, and the Evaluator object has only one in the global, configuration object for each HTTP probe.

> Just a different design style

haoel:
> IMHO, If I implemented the feature, I would prefer to splitting the Evaluator and Configuration into different objects, and the Evaluator object has only one in the global, configuration object for each HTTP prober.
> 
> > Just a different design style

I considered separating the Evaluator and Configuration, but they are the exactly same object which would introduce the duplication code we have to maintain their consistency.  And there is a Variable object inside, so finally, just use one struct for both of them.

If we only use one global Evaluator object for all HTTP probers, then we have to manage the different configurations for different HTTP probers, as they would have different document format(xml/html/json/txt) and different expressions. And this also would introduce multiple threads complexity. I prefer to have the dedicated Evaluator object for each HTTP probers.

localvar (approved):

zhao-kun (approved):

PR #197, opened 2022-08-17
Add more test cases

haoel:
Improvements
- Add test cases for the `channel`, `conf`, `report`, `probe` package
- Add test cases for the `ssh` and `http` probe
- Using go 1.19 for lint
- Using atomic for set/get `DryNotify` variable

### DIFF
--- a/channel/channel.go
+++ b/channel/channel.go
@@ -160,7 +160,7 @@ func (c *Channel) WatchEvent(wg *sync.WaitGroup) {
 			log.Infof("[%s / %s]: %s (%s) - Status changed [%s] ==> [%s]",
 				kind, c.Name, result.Name, result.Endpoint, result.PreStatus, result.Status)
 			for _, n := range c.Notifiers {
-				if GetDryNotify() == true {
+				if IsDryNotify() == true {
 					n.DryNotify(result)
 				} else {
 					go n.Notify(result)
--- a/channel/channel.go
+++ b/channel/channel.go
@@ -15,6 +15,7 @@
  * limitations under the License.
  */
 
+// Package channel implements the channel interface
 package channel
 
 import (
--- a/channel/channel.go
+++ b/channel/channel.go
@@ -159,7 +159,7 @@ func (c *Channel) WatchEvent(wg *sync.WaitGroup) {
 			log.Infof("[%s / %s]: %s (%s) - Status changed [%s] ==> [%s]",
 				kind, c.Name, result.Name, result.Endpoint, result.PreStatus, result.Status)
 			for _, n := range c.Notifiers {
-				if dryNotify {
+				if GetDryNotify() == true {
 					n.DryNotify(result)
 				} else {
 					go n.Notify(result)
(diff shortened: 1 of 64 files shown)


channel/manager.go
```diff
@@ -19,18 +19,28 @@ package channel
...
+	"sync/atomic"
 
 	"github.com/megaease/easeprobe/notify"
 	"github.com/megaease/easeprobe/probe"
 )
 
 var channel = make(map[string]*Channel)
 var wg sync.WaitGroup
-var dryNotify bool
+var dryNotify atomic.Value
```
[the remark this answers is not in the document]
haoel: At this moment, I think it's better to be compatible with go 1.18. As the go 1.19 is new, could have some bugs.

channel/channel.go
```diff
@@ -159,7 +160,7 @@ func (c *Channel) WatchEvent(wg *sync.WaitGroup) {
 			log.Infof("[%s / %s]: %s (%s) - Status changed [%s] ==> [%s]",
 				kind, c.Name, result.Name, result.Endpoint, result.PreStatus, result.Status)
 			for _, n := range c.Notifiers {
-				if dryNotify {
+				if GetDryNotify() == true {
```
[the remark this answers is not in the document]
haoel: I'd like to use this kind of style. it's  explicit to read. 

`if flag == true   { ... } `   is more readable then `if flag { ...}`
`if flag == false { ... }`  is better then `if !flag { ... }`

report/types_test.go
```diff
@@ -65,3 +65,10 @@ func TestFormatYAML(t *testing.T) {
 	testFormatYAML(t, SMS, "sms\n")
 	testFormatYAML(t, Unknown, "unknown\n")
 }
+
+func TestBadFormat(t *testing.T) {
+	buf := ([]byte)("- asdf::")
+	var f Format
+	err := yaml.Unmarshal(buf, &f)
+	assert.NotNil(t, err)
+}
```
zhao-kun, line 74: Why do we need this? it hasn't relevant to our codes

haoel: I don't understand your comment.  The `Format` is a type that could have marshal/unmarshal operations. it's relevant. and this test case just tests the bad yaml format that causes the unmarshal error.

zhao-kun: got it

channel/manager.go
```diff
@@ -19,18 +19,28 @@ package channel
...
+}
 
 // SetDryNotify sets the global dry run flag
 func SetDryNotify(dry bool) {
-	dryNotify = dry
+	dryNotify.Store(dry)
+}
+
+// GetDryNotify returns the dry run flag
+func GetDryNotify() bool {
```
zhao-kun, line 42: I think `IsDryNotify` is a better naming.

haoel: done!

localvar (approved):

zhao-kun (approved):

PR #198, opened 2022-08-22
custom host header

xiaomao87:
* https://pkg.go.dev/net/http#Request

	// For client requests, Host optionally overrides the Host
	// header to send. If empty, the Request.Write method uses
	// the value of URL.Host. Host may contain an international
	// domain name.
	Host string

### DIFF
--- a/probe/http/http.go
+++ b/probe/http/http.go
@@ -188,7 +188,11 @@ func (h *HTTP) DoProbe() (bool, string) {
 	}
 	req.Header.Set("User-Agent", global.OrgProgVer)
 	for k, v := range h.Headers {
-		req.Header.Set(k, v)
+		if strings.EqualFold(k, "host") {
+			req.Host = v
+		} else {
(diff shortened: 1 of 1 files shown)


probe/http/http.go
```diff
@@ -188,7 +188,11 @@ func (h *HTTP) DoProbe() (bool, string) {
 	}
 	req.Header.Set("User-Agent", global.OrgProgVer)
 	for k, v := range h.Headers {
-		req.Header.Set(k, v)
+		if k == "Host" {
```
zhao-kun, line 191: In the HTTP protocol [1]:
>Each header field consists of a name followed by a colon (":") and the field value. Field names are case-insensitive.

It's better to judge it in a case-insensitive way.

[1]: https://www.rfc-editor.org/rfc/rfc9110.html#name-field-names

haoel (approved):
Thanks for this PR

Issue #201, opened 2022-08-26
If there are many sites, can I include conf files like nginx?


feeops:
If there are many sites
For example, thousands of sites, all written in a config.yaml is too long and ugly

Can you include conf files like nginx?
That is, include files in a directory to make it easier to distinguish

haoel:
sorry,  currently, the YAML does not include any kind of "import" or "include" statement.

and I think we can support the multiple yaml files. but I need time to think how to implement it.,

haoel:
P.S. I need to evaluate these two solutions: 1) develop a YAML editor UI, 2) support the multiple YAML files

the first one would address the pains of YAML configuration much more than the second one, but the work efforts are much more.

feeops:
> P.S. I need to evaluate these two solutions: 1) develop a YAML editor UI, 2) support the multiple YAML files
> 
> the first one would address the pains of YAML configuration much more than the second one, but the work efforts are much more.

知名时序数据库VictoriaMetrics是采用后者，可以加载多个yaml文件
https://github.com/VictoriaMetrics/VictoriaMetrics/blob/master/docs/vmagent.md#loading-scrape-configs-from-multiple-files

haoel:
> > P.S. I need to evaluate these two solutions: 1) develop a YAML editor UI, 2) support the multiple YAML files
> 
> > 
> 
> > the first one would address the pains of YAML configuration much more than the second one, but the work efforts are much more.
> 
> 
> 
> 知名时序数据库VictoriaMetrics是采用后者，可以加载多个yaml文件
> 
> https://github.com/VictoriaMetrics/VictoriaMetrics/blob/master/docs/vmagent.md#loading-scrape-configs-from-multiple-files

Would you like to contribute this PR?

feeops:
> 

Sorry.My program skill is not good enough.

yafengstark:
> 
I can hava a try. Give me some time, Please.

PR #215, opened 2022-09-16
Support the multiple YAML files

yafengstark:
#201 
I add some code to support the multiple YAML files. Now you can use multiple `-f` to split configuration. 
I am a new Golang user. Please review the code carefully.
Thanks.

### DIFF
--- a/cmd/easeprobe/main.go
+++ b/cmd/easeprobe/main.go
@@ -75,10 +75,18 @@ func main() {
 	////////////////////////////////////////////////////////////////////////////
 
 	dryNotify := flag.Bool("d", os.Getenv("PROBE_DRY") == "true", "dry notification mode")
-	yamlFile := flag.String("f", getEnvOrDefault("PROBE_CONFIG", "config.yaml"), "configuration file")
+
+	var yamlFiles conf.FFlags
+	flag.Var(&yamlFiles, "f", "configuration file")
+
 	jsonSchema := flag.Bool("j", false, "show JSON schema")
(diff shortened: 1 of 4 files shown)


haoel:
@yafengstark thanks so much for this PR. and sorry for the delayed response.

I appreciate your work. And I think we should design this carefully before we start to code. 

the following things we need to think about:

1) if there has some configuration conflict, how to solve the conflict? 
2) do you think we can just simply concatenate all of the yaml files to a single yaml file, then we can keep the rest thing with no change?
3) if 2) is the better solution, why not just use a shell helper script to merge the multiple yaml files to a single yaml? 

Overall, actually, this problem is how to manage the EaseProbe's configuration, I think we can do this outside of EaseProbe and keep EaseProbe simple and stupid.

yafengstark:
> 

@haoel thanks for your reply. Sorry for the delayed response.

You point out the key. I agree with you that we should do this outside of EaseProbe.

I find the repository [https://github.com/herrjulz/aviator](aviator).The repository  specifies the path to the base YAML file. All other YAML files will be merged on top of this YAML file.

I do some simple tests, it works.
But I do not kown When some configuration conflict will occur, so I do not test the conflict.

What do you think of this repository？If there is anything to be done, I can do some work.

thanks.

haoel:
the `aviator` looks not bad.  we probably can introduce the `conf.d` for the pieces of yaml files, and the original `config.yaml` is the base yaml.

haoel:
This PR has been addressed by PR #218

PR #232, opened 2022-10-13
Add EASEPROBE_TIME to shell env

### LINKED ISSUE #231: EASEPROBE_TIME is only provided in JSON format
**Environment (please complete the following information):**
 - OS: Linux VM-8-5-centos 3.10.0-1160.76.1.el7.x86_64 # 1 SMP Wed Aug 10 16:21:17 UTC 2022 x86_64 x86_64 x86_64 GNU/Linux
 - EaseProbe Version EaseProbe v1.7.0 @ a4255478e017

**Describe the bug**
When looking at shell notification script, values are both provided by JSON format or env variables, however they're misaligned. `EASEPROBE_TIME` is part of JSON format but not in env variables.

**To Reproduce**
1. set shell script as https://github.com/megaease/easeprobe/blob/main/resources/scripts/notify/notify.sh
2. see output

**Expected behavior**
`EASEPROBE_TIME` env variable should also exist.

The reason I need this is because, the docker image was built based on alpine so there's quite a little utitlites avaiable out of box, so I cannot run `jq` without a self-made image. `EASEPROBE_TIMESTAMP` is in env variable but it's a millisecond epoch so `date -d` is not able to calc it unless I do the math first.

zhzy0077:
fixes #231

### DIFF
--- a/docs/Manual.md
+++ b/docs/Manual.md
@@ -1215,6 +1215,7 @@ notify:
   #  - EASEPROBE_NAME: probe name
   #  - EASEPROBE_STATUS: "up" or "down"
   #  - EASEPROBE_RTT: round trip time in milliseconds
+  #  - EASEPROBE_TIME: time of probe time(formatted by timeformat configured in settings section)
   #  - EASEPROBE_TIMESTAMP: timestamp of probe time
   #  - EASEPROBE_MESSAGE: probe message
   # and offer two formats of string
--- a/docs/Manual.md
+++ b/docs/Manual.md
@@ -1221,7 +1221,7 @@ notify:
   # and offer two formats of string
   #  - EASEPROBE_JSON: the JSON format
   #  - EASEPROBE_CSV: the CSV format
-  # The CVS format would be set for STDIN for the shell command.
+  # The CSV format would be set for STDIN for the shell command.
   # (see the example: resources/scripts/notify/notify.sh)
   shell:
     - name: "shell alert service"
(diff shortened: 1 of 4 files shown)


report/result.go
```diff
@@ -290,6 +290,7 @@ func ToShell(r probe.Result) string {
 	env["EASEPROBE_ENDPOINT"] = r.Endpoint
 	env["EASEPROBE_STATUS"] = r.Status.String()
 	env["EASEPROBE_TIMESTAMP"] = fmt.Sprintf("%d", r.StartTimestamp)
+	env["EASEPROBE_TIME"] = FormatTime(r.StartTime)
```
zhzy0077, line 293: I'm not quite sure about the environment name as we're not able to easily change it in the future (breaks the backwards comaptibitliy of existing shell scripts) 
Need double check

haoel: I think the name is fine. 

could you please update the demo shell script and the document?

zhzy0077: Great. Updated.

docs/Manual.md
```diff
@@ -1215,12 +1215,13 @@ notify:
...
   #  - EASEPROBE_STATUS: "up" or "down"
   #  - EASEPROBE_RTT: round trip time in milliseconds
+  #  - EASEPROBE_TIME: time of probe time(formatted by timeformat configured in settings section)
   #  - EASEPROBE_TIMESTAMP: timestamp of probe time
   #  - EASEPROBE_MESSAGE: probe message
   # and offer two formats of string
   #  - EASEPROBE_JSON: the JSON format
   #  - EASEPROBE_CSV: the CSV format
-  # The CVS format would be set for STDIN for the shell command.
+  # The CSV format would be set for STDIN for the shell command.
```
zhzy0077, line 1224: I piggyback a typo fix.

haoel (approved):

PR #235, opened 2022-10-17
bug-fixing: http probe report the unsupported document type

haoel:
If http probe didn't configure the `eval/doc`,  EaseProbe would report the error - `Unsupported document type: unsupported`

This bug only can be reproduced after PR #192 

here is the fix - check if the `eval` configuration is not configured, then skip the eval configration.


samanhappy (approved):
LGTM

proditis (approved):
I can confirm this makes the error message go away :+1:

PR #236, opened 2022-10-19
Minutely sla reports

proditis:
This PR includes the following:

* [ ] Add minutely SLA schedule interval
* [ ] Add missing code for "hourly" SLA schedule interval
* [ ] Minor updates to wording of the messages
* [ ] Remove a duplicate `log.Infof`

PR #237, opened 2022-10-20
Optimizing the SLA Report

haoel:
The PR is the continuous work for  PR #236, which optimize the following works:

- [x] make the `minutely` schedule at exact `mm:00`
- [x] make the `hourly` schedule at  exact `hh:mm:00`
- [x] set the gocron timezone as configured timezone instead of hard code UTC. such as: `timezone:  "Asia/Shanghai"` in config.yaml, so that the log message would output the configured timezone for `NextRun()`.
- [x] restore a log message which would log the `NextRun()` time when EaseProbe start.

### DIFF
--- a/cmd/easeprobe/report.go
+++ b/cmd/easeprobe/report.go
@@ -23,6 +23,7 @@ import (
 	"github.com/go-co-op/gocron"
 	"github.com/megaease/easeprobe/channel"
 	"github.com/megaease/easeprobe/conf"
+	"github.com/megaease/easeprobe/global"
 	"github.com/megaease/easeprobe/probe"
 	log "github.com/sirupsen/logrus"
 )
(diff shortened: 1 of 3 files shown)


cmd/easeprobe/report.go
```diff
@@ -116,4 +117,7 @@ func scheduleSLA(probers []probe.Prober) {
 	}
 
 	cron.StartAsync()
+
+	_, t := cron.NextRun()
+	log.Infof("The SLA report will be schedule at %s", t.Format(conf.Get().Settings.TimeFormat))
```
proditis, line 122: Do we need this back ? I removed it on the previous PR as it was displaying the same message 2 times. One was from this line and the other from line 94

proditis (commented):
LGTM other than the addition of `log.Infof`

cmd/easeprobe/report.go
haoel: Sorry, I misunderstood at that time. 

This  different is 
- here, the log.Infof() is for the first time when EaseProbe starts. 
- line 94 will be executed after the SLA  report is sent. 

So, without this log, we cannot see when the first SLA report will be sent.

proditis (approved):
LGTM

PR #245, opened 2022-11-02
Enhance email notification: add an ssl option

xiaoxuanzi:
Hi, I use Email notification and find the following error message:
> [email / test-dev / Notification] Retried to send 3/3 - tls: first record does not look like a TLS handshake

I think it's better to provide an option and let the user choose.

### DIFF
--- a/notify/email/email.go
+++ b/notify/email/email.go
@@ -19,16 +19,15 @@
 package email
 
 import (
-	"crypto/tls"
-	"fmt"
 	"net"
-	"net/smtp"
+	"strconv"
 	"strings"
 
 	"github.com/megaease/easeprobe/global"
 	"github.com/megaease/easeprobe/notify/base"
 	"github.com/megaease/easeprobe/report"
 	log "github.com/sirupsen/logrus"
+	"gopkg.in/gomail.v2"
 )
 
 // NotifyConfig is the email notification configuration
@@ -54,7 +53,12 @@ func (c *NotifyConfig) Config(gConf global.NotifySettings) error {
 // SendMail sends the email
 func (c *NotifyConfig) SendMail(subject string, message string) error {
 
-	host, _, err := net.SplitHostPort(c.Server)
+	host, p, err := net.SplitHostPort(c.Server)
+	if err != nil {
+		return err
+	}
+
+	port, err := strconv.Atoi(p)
 	if err != nil {
 		return err
 	}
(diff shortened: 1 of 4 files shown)


notify/email/email.go
```diff
@@ -76,8 +77,25 @@ func (c *NotifyConfig) SendMail(subject string, message string) error {
...
 
+	if !c.SSL {
+		msg := []byte(body)
+		err = smtp.SendMail(c.Server, auth, c.User, recipients, msg)
+		if err != nil {
+			return err
+		}
+
+		return nil
+	}
```
samanhappy, line 97: Can we just use `client, err := smtp.Dial(c.Server)` when ssl is false, so the code logic can be keep same?

xiaoxuanzi: That's what I did at first, but many details need to be handled, and encountered some mistakes.
I found The function SendMail provided by smtp implements this logic, so we don't need to repeat it.

haoel: so, do you mean the failure of  `smtp.SendEmail()` would impact the `smtp.NewClient()` and the rest SMTP operations?

xiaoxuanzi: I mean, if I use `client, err := smtp.Dial(c.Server)` when ssl is false, I need to deal with some details.
For example, I need to use `err = c.Hello()` etc.
If I ignore these, it will report an error.
This is inconvenient. I think it is a good choice to use SendMail provided by smtp directly.

haoel:
@xiaoxuanzi Thanks for this contribution. however, the Unit Test  - TestEmail() failed, 
 https://github.com/megaease/easeprobe/actions/runs/3374505121/jobs/5617127318#step:5:483

Please help to fix that!

xiaoxuanzi:
> @xiaoxuanzi Thanks for this contribution. however, the Unit Test - TestEmail() failed, https://github.com/megaease/easeprobe/actions/runs/3374505121/jobs/5617127318#step:5:483
> 
> Please help to fix that!

Hi @haoel.
Fixed, please review it.

notify/email/email.go
```diff
@@ -76,8 +77,21 @@ func (c *NotifyConfig) SendMail(subject string, message string) error {
...
+		return r == ';' || r == ','
+	}
+
+	recipients := strings.FieldsFunc(c.To, split)
+
 	auth := smtp.PlainAuth("", c.User, c.Pass, host)
 
+	if !c.SSL {
+		msg := []byte(body)
+		err = smtp.SendMail(c.Server, auth, c.User, recipients, msg)
```
samanhappy, line 91: According to the comments, SendMail can switch to TLS if possible, can we just use it to support TLS and pure TCP at the same time?

xiaoxuanzi:
I did some tests.

Configure：
```
  email:
    - name: "test-port-25"
      server: smtp.qq.com:25
      username: xxxx
      password: xxxx
      to: "xxx"
    - name: "test-port-465"
      server: smtp.qq.com:465
      username: xxxx
      password: xxxx
      to: "xxx"
    - name: "test-port-587"
      server: smtp.qq.com:587
      username: xxxx
      password: xxxx
      to: "xxx"
```

Log:

```
INFO[2022-11-09T13:09:42+08:00] [email / test-port-465 / Notification] - tcp-test Failure - successfully sent!
INFO[2022-11-09T13:09:42+08:00] [email / test-port-25 / Notification] - tcp-test Failure - successfully sent!
INFO[2022-11-09T13:09:42+08:00] [email / test-port-587 / Notification] - tcp-test Failure - successfully sent!
```
Received:
image

haoel (approved):
@xiaoxuanzi thanks for the test report, looks very good!

PR #272, opened 2023-01-11
Optimizing and adding new SLA prometheus  metrics

haoel:
- Optimizing the SLA total metrics
- Adding the SLA total time metrics

### DIFF
--- a/docs/Manual.md
+++ b/docs/Manual.md
@@ -562,7 +562,7 @@ EaseProbe supports Prometheus metrics exporter. The Prometheus endpoint is `http
 
 Currently, All of the Probers support the following metrics:
 
-  - `total`: the total probe number of probes
+  - `total`: the total number of probes
   - `total_time`: the total time(seconds) of status up or down
   - `duration`: Probe duration in milliseconds
   - `status`: Probe status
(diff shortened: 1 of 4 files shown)


probe/base/metrics.go
```diff
@@ -25,18 +25,21 @@ import (
 
 // metrics is the probe metrics
 type metrics struct {
-	Total    *prometheus.CounterVec
-	Duration *prometheus.GaugeVec
-	Status   *prometheus.GaugeVec
-	SLA      *prometheus.GaugeVec
+	TotalCnt  *prometheus.GaugeVec
```
zhao-kun, line 28: It is suggested not to change the name of this metric due to compatibility issues. In considering compatibility issues, the names of metrics that are already in use should not be changed. If the name of a metric is changed, it may cause problems in existing code, documentation, tools, or systems. This can result in errors, confusion, or other issues, so the impact should be carefully considered before making any changes to the name

docs/Manual.md
```diff
@@ -562,7 +562,8 @@ EaseProbe supports Prometheus metrics exporter. The Prometheus endpoint is `http
 
 Currently, All of the Probers support the following metrics:
 
-  - `total`: the total number of probes
+  - `total`: the total probe number of probes
```
proditis, line 565:
```suggestion
  - `total`: the total number of probes
```

zhao-kun (approved):

samanhappy (approved):

PR #294, opened 2023-03-01
chore(deps): replace go-ping with pro-bing

haoqixu:
[go-ping/ping](https://github.com/go-ping/ping) is no longer maintained. Switch to the recommended, actively maintained fork: https://github.com/prometheus-community/pro-bing

https://github.com/go-ping/ping/issues/221


haoel (approved):
Thanks you for this migration!

PR #295, opened 2023-03-01
fix: use buffered os.Signal channel as argument to signal.Notify

haoqixu:
Use buffered os.Signal channel as argument to signal.Notify to ensure that we do not drop a signal notification.

### DIFF
--- a/cmd/easeprobe/main.go
+++ b/cmd/easeprobe/main.go
@@ -209,7 +209,7 @@ func main() {
 	////////////////////////////////////////////////////////////////////////////
 	//                              Graceful Shutdown                         //
 	////////////////////////////////////////////////////////////////////////////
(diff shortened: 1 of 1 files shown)


haoel:
Thanks for this PR, cloud you please let me understand the situation which could drop the signal notification?

haoqixu:
https://pkg.go.dev/os/signal#Notify:
> Package signal will not block sending to c: the caller must ensure that c has sufficient buffer space to keep up with the expected signal rate. For a channel used for notification of just one signal value, a buffer of size 1 is sufficient. 

Any signal which arrives after `signal.Notify` but before `<-done` will get discarded. It is just technically possible and not a big problem here.

haoel (approved):

haoel:
Thanks for this explanation

Issue #385, opened 2023-07-21
Add constant labels for Prometheus metrics


qdongxu:
**Background & X Problem**

**EaseProbe** probes a large number of endpoints, but currently labels can only be configured in the Prometheus scrape_configs, this can only label the probe itself.  

**Proposal and Expectation**
 It's expected that EaseProble behaves as a local exporter on the target node and the labels specific to the target nodes. 

**Solutions & Y Problems**

Propose to add a  `labels` item to each probe module, which contains constant labels. Here's a example, adding constant labels to an HTTP prober.

```yaml
   http:
     - name: Web Service
       url: http://example.com:1080
       labels:
          service: web_example
          env: wg-idc-prod
       alert:
         strategy: regular
         factor: 
```
the corresponding Prometheus metric looks like:

> EaseProbe_http_status{env="wg-idc-prod", endpoint="http://example.com:1080", instance="127.0.0.1:8181", job="easeprobe", name="Web Service", service="web_example"}

**Additional context**

PR #386, opened 2023-07-21
add constant labels for the metrics (#385)

qdongxu:


qdongxu:
Looks like there's an issue in `gopkd.in/yaml.v3@v3.0.1` . 

with config:
```yaml
http: # http probes
  - name: EaseProbe Github
    url: https://github.com/megaease/easeprobe
    labels:
      abc: xx
      def: y
      instance: github.com/megaeas
  - name: 163
    url: https://www.163.com
```

the lablels of 'EaseProbe Github' binds to '163' as well:

> EaseProbe_http_status{abc="xx", def="y", endpoint="https://github.com/megaease/easeprobe", exported_instance="github.com/megaeas", instance="localhost:8181", job="easeprobe", name="EaseProbe Github"}
1
> EaseProbe_http_status{abc="xx", def="y", endpoint="https://www.163.com", exported_instance="github.com/megaeas", instance="localhost:8181", job="easeprobe", name="163"}
1

qdongxu:
After changing the order:
```yaml
http: # http probes
  - name: 163
    url: https://www.163.com
  - name: EaseProbe Github
    url: https://github.com/megaease/easeprobe
    labels:
      abc: xx
      def: y
      instance: github.com/megaeas
```
the labels was overwritten into empty:

> EaseProbe_http_status{endpoint="https://github.com/megaease/easeprobe", instance="localhost:8181", job="easeprobe", name="EaseProbe Github"}
1
> EaseProbe_http_status{endpoint="https://www.163.com", instance="localhost:8181", job="easeprobe", name="163"}
1

qdongxu:
This was caused by the strategy of `Prometheus`, the Registry requires each metric ( identified by the metric name) should have the same label names. https://github.com/prometheus/client_golang/issues/155 . 

that is the config should be as below:

```yaml
http: # http probes
  - name: 163
    url: https://www.163.com
    labels:
      abc: x163
      def: y163
      instance: 163
  - name: EaseProbe Github
    url: https://github.com/megaease/easeprobe
    labels:
      abc: xx
      def: y
      instance: megaease
```

but since the golang map is unordered, the configuration should be like below:

```yaml
http: # http probes
  - name: 163
    url: https://www.163.com
    labels:
      - name: abc
         value: x163
     - name: def
       value: y163
    - name: instance
      value: 163
  - name: EaseProbe Github
    url: https://github.com/megaease/easeprobe
    labels:
      - name: abc
         value: xx
     - name: def
       value:  y
    - name: instance
      value: megaease
```

this looks verbose and error-prone.

samanhappy:
Thanks for contributing this PR, there is a week from last commit, is it ready for review now?

qdongxu:
The solution is a little ugly. As per the Prometheus Collector constraint, a metric must have the same set of constant labels, which looks like below, all prober should define the three labels: `abc`, `def`, `instance`:

```
http: # http probes
  - name: 163
    url: https://www.163.com
    labels:
      - name: abc
         value: x163
     - name: def
       value: y163
    - name: instance
      value: 163
  - name: EaseProbe Github
    url: https://github.com/megaease/easeprobe
    labels:
      - name: abc
         value: xx
     - name: def
       value:  y
    - name: instance
      value: megaease
```
I've tried to use a collector for each prober but didn't find a way to bind all the collectors to one HTTP handler. 

Anyway, this meets my reqirement and I am deploying it to the production environment. The code is ready for review if this solution is acceptable.

samanhappy:
You're correct. The current solution is overly verbose and susceptible to errors. We should strive to discover a more effective approach.

Is it possible to consolidate these customized labels into a singular Prometheus label field? This would enable us to maintain label consistency.

However, I'm uncertain about the ease of management within Prometheus consumers like Kindna. What are your thoughts on this?

localvar:
this solution requires all metrics to have the same set of constant labels, but users may forget this requirement and get an unexpected result.

is it possible that we merge the constant labels of all metrics to one set, and set the value of a label to empty if a metric does not need it while keep the configured value if the metric need it?

qdongxu:
This PR is ready for review. the merge solution works fine. it passes the collector's checking and the empty labels are discarded by the exporter.

docs/Manual.md
```diff
@@ -273,7 +278,11 @@ http:
   # A Website
   - name: MegaEase Website (Global)
     url: https://megaease.com
-
+    labels:
+      - name: team
+        value: ease
+      - name: owner
+        value: megaease
```
localvar, line 285: as we merged all constant labels into one set, using a map instead of an array will be more user friendly.

```suggestion
    labels:
      team: ease
      owner: megaease
```

qdongxu: updated

qdongxu:
@samanhappy @localvar this PR is ready for review now.

conf/conf.go
```diff
@@ -436,7 +436,9 @@ func isProbe(t reflect.Type) bool {
 // AllProbers return all probers
 func (conf *Conf) AllProbers() []probe.Prober {
 	log.Debugf("--------- Process the probers settings ---------")
-	return allProbersHelper(*conf)
+	all := allProbersHelper(*conf)
+	MergeConstLabels(all)
```
localvar, line 440: `AllProbers` may be called more than once (even it is not currently, the name suggest it could be), which result in `MergeConstLabels` be called more than once, this is not necessary and may cause a bug in the future.

propose moving `MergeConstLabels` to somewhere that will only be called once.

conf/constlables.go
```diff
@@ -0,0 +1,34 @@
+package conf
+
+import (
+	"github.com/megaease/easeprobe/probe"
+)
+
+var constLabels = make(map[string]bool)
```
localvar, line 7: why it is a global variable? I think we can make it a local variable by refactor the code below, this could make the code more maintainable and avoid potential bugs.

metric/prometheus.go
```diff
@@ -179,3 +194,10 @@ func RemoveInvalidChars(name string) string {
 	}
 	return string(result)
 }
+
+func AddConstLabels(labels prometheus.Labels, constLabels prometheus.Labels) prometheus.Labels {
+	for k, v := range constLabels {
+		labels[k] = v
+	}
```
localvar, line 201: should check if a label exists both in `labels` and `constLabels` and report the error.

qdongxu: updated

conf/constlables.go
qdongxu: this was a digress of an original design. updated.

conf/conf.go
qdongxu: updated.

localvar (approved):
I'll approve this PR now, but please fix the github workflow errors.

qdongxu:
fixed. but I cannot verify fully as some unit tests failed on mac m1 due to the bou.ke/monkey issue. please run a github workflow to verify it.

conf/constlables.go
```diff
@@ -0,0 +1,52 @@
...
+	"github.com/megaease/easeprobe/probe"
+)
+
+// MergeConstLabels merge const labels from all probers.
+//
+//	Prometheus requires all metric  of the same name have the same set of labels in a collector
+func MergeConstLabels(ps []probe.Prober) {
+	var constLabels = make(map[string]bool)
+	for _, p := range ps {
+		for k, _ := range p.LabelMap() {
```
samanhappy, line 30: please fix  this and the following code lint warnings

metric/prometheus.go
```diff
@@ -75,23 +77,25 @@ func NewCounter(namespace, subsystem, name, metric string,
 			Name: metricName,
 			Help: help,
 		},
-		labels,
+		mergeLabels(labels, constLabels),
 	)
-	prometheus.MustRegister(counterMap[metricName])
 
+	// registries[len(registries)-1].MustRegister(m)
```
samanhappy, line 83:
```suggestion
```

samanhappy (approved):
LGTM，please fix all the code lint warnings and some verbose comments

metric/prometheus.go
qdongxu: updated.

conf/constlables.go
qdongxu: updaed.

conf/constlables.go
```diff
@@ -0,0 +1,52 @@
...
+
+package conf
+
+import (
+	"github.com/megaease/easeprobe/probe"
+)
+
+// MergeConstLabels merge const labels from all probers.
+//
+//	Prometheus requires all metric  of the same name have the same set of labels in a collector
```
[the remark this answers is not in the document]
qdongxu: updated

zhao-kun (approved):

PR #388, opened 2023-07-27
bugfix:strategy type error

uliian:
The type definition of the "strategy" field in the schema.json is incorrect. It should be defined as a string type, but it is currently defined as an integer.

### DIFF
--- a/resources/schema.json
+++ b/resources/schema.json
@@ -415,7 +415,8 @@
     "global_NotificationStrategySettings": {
       "properties": {
         "strategy": {
-          "type": "integer",
+          "type": "string",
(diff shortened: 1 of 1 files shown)


localvar (approved):

samanhappy:
Hi uliian,

Thanks for submitting this pull request. Just a heads up - the schema.json file is auto-generated by a GitHub action, so we don't modify it directly.

I went ahead and submitted PR #388 that updates the schemajson of the strategy field. Would you mind taking a look when you have a chance? Let me know if you have any other questions!

localvar:
this file is auto generated and should not be modified directly, the issue is fixed via #389 .

PR #401, opened 2023-08-16
fix unix test compiling failed on MAC M1

qdongxu:
No any unit test not be executed as the compiling failed with this issue :

https://stackoverflow.com/questions/74112738/underfined-function-error-in-go-jmptofunctionvalue
https://github.com/jenkins-x/jx/issues/2081
https://github.com/procodr/monkey 

the monkey patch unit tests still failed but the compile finished anyway and unit tests that are not related to the monkey patch can be executed.

### DIFF
--- a/conf/conf_test.go
+++ b/conf/conf_test.go
@@ -28,7 +28,6 @@ import (
 	"testing"
 	"time"
 
-	"bou.ke/monkey"
 	"github.com/megaease/easeprobe/global"
 	"github.com/megaease/easeprobe/notify"
 	"github.com/megaease/easeprobe/notify/discord"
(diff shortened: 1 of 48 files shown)


samanhappy:
The original monkey repository has been archived by the owner, so we need to find a suitable alternative, the repository you are using seems too temporary for me, how about this [gomonkey](https://github.com/agiledragon/gomonkey) repository? Could you help to test it?

qdongxu:
Testing gomonkey on MacOS M1,  the issues https://github.com/agiledragon/gomonkey/issues?q=is%3Aissue++Macos+m1 still exists.  and the interface and method of use changed in an incompatible way. It would only be worth doing the refactor until gomonkey is stable on MacOS M1/M2.

samanhappy:
You are right, gomonkey is not stable now, we should wait or find another library, thanks for you work.

PR #433, opened 2023-10-27
Update Manual.md

jiacheo:
update yunpian doc link

### DIFF
--- a/docs/Manual.md
+++ b/docs/Manual.md
@@ -1097,7 +1097,7 @@ Support SMS notification with multiple SMS service providers
 
 - [Twilio](https://www.twilio.com/sms)
 - [Vonage(Nexmo)](https://developer.vonage.com/messaging/sms/overview)
-- [YunPian](https://www.yunpian.com/doc/en/domestic/list.html)
+- [YunPian](https://www.yunpian.com/official/document/sms/en/domestic_list)
 
 The plugin supports the following parameters:
  - `name`: A unique name for this notification endpoint
--- a/docs/Manual.md
+++ b/docs/Manual.md
@@ -1097,7 +1097,7 @@ Support SMS notification with multiple SMS service providers
 
 - [Twilio](https://www.twilio.com/sms)
 - [Vonage(Nexmo)](https://developer.vonage.com/messaging/sms/overview)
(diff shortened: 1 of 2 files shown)


samanhappy (approved):

docs/Manual.md
```diff
@@ -1097,7 +1097,7 @@ Support SMS notification with multiple SMS service providers
 
 - [Twilio](https://www.twilio.com/sms)
 - [Vonage(Nexmo)](https://developer.vonage.com/messaging/sms/overview)
-- [YunPian](https://www.yunpian.com/doc/en/domestic/list.html)
+- [YunPian](https://www.yunpian.com/official/document/sms/en/domestic_list)
```
suchen-sci, line 1100: Hi, thanks for you contribution. I think it would be better to put both English and Chinese doc here. Could you please add it?

jiacheo: Sure, but I think it's not that necessary.

According to YunPian's doc website, it renders default in Chinese. If we want an English version, we can add a 'lang=en' pramameter in the url or hover the `LANGUAGE DROPDOWN LIST` at the top-right screen and select the language we want.

> https://www.yunpian.com/official/document/sms/en/domestic_list?lang=en  

Just like the above links of `Twilio` and `Vonage`, they render default in English and could be changed by selecting the specific language at the top-right screen too.

suchen-sci: Then maybe use english version of doc? 

The `Manual.md` is written in english. So I am afraid this Chinese doc may be confused for some users who don't know about Chinese. Consider you open a website, and it is written all by Finnish language, it actually hard for you to find the position to switch language back to Chinese. 

Since YunPian provides english doc, and this manual is written in english. I think maybe the english version of doc is a little bit better here. `https://www.yunpian.com/doc/en/domestic/list.html` the previous link is english version too. 
I am sure Chinese users who can read this manual can also read the english version of doc. I believe in long term, link english version of doc may bring some foreign users for YunPian too!

What do you think?

jiacheo: I got you, will update it.

jiacheo: I also updated the link in `README.md` this time. @suchen-sci

suchen-sci (approved):

PR #473, opened 2024-01-12
log error when meet invalid json

suchen-sci:
In some special cases, json may failed to escape. This pr log errors in these cases and help user to find bug quickly.


samanhappy (approved):
this is a useful optimization

PR #483, opened 2024-02-27
Fix ssh timeout

suchen-sci:


probe/ssh/ssh.go
```diff
@@ -259,11 +260,20 @@ func (s *Server) RunSSHCmd() (string, error) {
...
 	session.Stdout = &stdoutBuf
 	session.Stderr = &stderrBuf
-	if err := session.Run(env + global.CommandLine(s.Command, s.Args)); err != nil {
-		return stderrBuf.String(), err
-	}
 
-	return stdoutBuf.String(), nil
+	errCh := make(chan error, 1)
+	go func() {
+		errCh <- session.Run(env + global.CommandLine(s.Command, s.Args))
```
LokiWager, line 266: Will this function exit due to a timeout?

suchen-sci: yeah, the `session.Run` may not exit although we call defer to close session and client. To make sure `session.Run` exit, i just add command to send signal when context timeout.

LokiWager (approved):

PR #533, opened 2024-06-04
fix: metric EaseProbe_ping_loss Add by mistake

fangpsh:
- gauge metric should be SET value, not ADD.
 - fix MaxRtt metric loss.


fangpsh:
I have found the commit that caused this bug, maybe we should  re-review this pull request：

https://github.com/megaease/easeprobe/pull/386/files#diff-2905b6c764234f33b4bf658a5c5e0fa0d82289c07f56455c6ed348e60c296c36L142

samanhappy:
Thanks for your contribution. These are indeed mistakes that were introduced in the previous PR.

samanhappy (approved):

PR #557, opened 2024-08-27
Make linger configurable

proditis:
The following PR is an attempt to make SO_LINGER option configurable. 

It introduces a new `bool` option named `nolinger` for tcp, tls, http and ssh probes which disables `SO_LINGER`. Default value set to `false` to match the previous easeprobe operation and not require configuration changes when updated.

Documentation also updated to include the addition of this option.

Notes:
* Needs testing (i am unable to test it currently)
* I dont know how to implement the new option for our test cases.

### DIFF
--- a/docs/Manual.md
+++ b/docs/Manual.md
@@ -277,7 +277,7 @@ http:
   # A Website without SO_LINGER
   - name: MegaEase Website (Global)
     url: https://megaease.com
-    linger: false
+    nolinger: true
     labels:
       team:  ease
       owner: megaease
@@ -346,7 +346,7 @@ http:
       expression: "x_time('//feed/updated') > '2022-07-01'" # the expression to evaluate.
     # configuration
     timeout: 10s # default is 30 seconds
-    linger: false # Do not set SO_LINGER
+    nolinger: true # Do not set SO_LINGER
 ```
 
 > **Note**:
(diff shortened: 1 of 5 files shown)


samanhappy (approved):
LGTM

suchen-sci:
Hi @proditis, 

Thank you for your pull request. However, there is a concern with this change. Prior to this PR, the default behavior was equivalent to setting `linger` to `true`. Although you've used `jsonschema` to set the default value of `linger` to `true`, this doesn't actually work in this case because we're using `yaml.Unmarshal` for loading, which doesn't enforce those defaults.

To maintain the previous default behavior, I strongly recommend using `nolinger` instead of `linger`. This would ensure that the behavior remains consistent.

suchen-sci (commented):
Hi @proditis,

Thank you for your pull request. However, there is a concern with this change. Prior to this PR, the default behavior was equivalent to setting `linger` to `true`. Although you've used `jsonschema` to set the default value of `linger` to true, this doesn't actually work in this case because we're using `yaml.Unmarshal` for loading, which doesn't enforce those defaults.

To maintain the previous default behavior, I strongly recommend using `nolinger` instead of `linger`. This would ensure that the behavior remains consistent.

suchen-sci (approved):

Mr-YYM:
when to release a new version? 😄

suchen-sci:
@Mr-YYM There is a new PR in progress, if that one is merged. We will release a new version.

PR #570, opened 2024-09-05
replace bou.ke/monkey with bytedance/mockey

### LINKED ISSUE #519: Find an alternative solution for bou.ke/monkey
Currently, all tests will fail with the following log:

```
# bou.ke/monkey
Error: ../../../go/pkg/mod/bou.ke/monkey@v1.0.2/replace.go:24:14: undefined: jmpToFunctionValue
```

This seems to be related to the library's implementation on Mac M1. The `bou.ke/monkey` library has been archived for a long time and has license restrictions as mentioned in [this issue](https://github.com/megaease/easeprobe/issues/216).

I've found an alternative library, [gomonkey](https://github.com/agiledragon/gomonkey), but there seems to be some API incompatibilities like `Patch`, `UnpatchAll`, etc. 

Any suggestions @suchen-sci ?

samanhappy:
As mentioned in #519, the `bou.ke/monkey` library has been archived for a long time and fails on the `macos-arm64` architecture. Consequently, we need to find an alternative library.

After extensive testing, I found that ~~`gomonkey`~~ `bytedance/mockey` is the only library that works, although its APIs are not compatible with `bou.ke/monkey`. To address this, I have introduced a separate monkey module in EaseProbe that encapsulates the ~~`gomonkey`~~ `bytedance/mockey` library. This approach minimizes the cost of modifying test code and allows us to easily replace the library in the future without affecting our test code.

Key code changes include:

1. **Makefile Update**: Changed `-gcflags=-l` to `-gcflags=all=-l` to accommodate changes for ~~`gomonkey`~~ `bytedance/mockey`.
2. **Test Files Update**: Most test files only required changing the import path from `bou.ke/monkey` to `github.com/megaease/easeprobe/monkey`. However, in `probe/ssh/ssh_test.go`, where the original usage was unsupported, I modified the logic accordingly.
3. **Monkey Module**: In `monkey/monkey.go`, I used a `sync.Map` to store all patches, which supports reset operations. A notable method called `wait` employs busy waiting to prevent test failures on `macos-arm64`. The exact cause of these errors is unclear to me, and this is a workaround. If anyone has a better solution, please let me know, as I am not very familiar with the ARM64 architecture.

Resolve #519

### DIFF
--- a/.github/workflows/test.yaml
+++ b/.github/workflows/test.yaml
@@ -27,7 +27,7 @@ jobs:
     runs-on: ${{ matrix.os }}
     strategy:
       matrix:
-        os: [ubuntu-latest, windows-latest, macos-13]
+        os: [ubuntu-latest, windows-latest, macos-latest]
     steps:
       - name: Checkout Source Code
         uses: actions/checkout@v4
--- a/Makefile
+++ b/Makefile
@@ -35,7 +35,7 @@ ${TARGET}: ${SOURCE}
 build: all
 
 test:
-	go test -gcflags=-l -cover -race ${TEST_FLAGS} -v ./...
+	go test -gcflags=all=-l -cover -race ${TEST_FLAGS} -v ./...
 
 docker:
 	sudo DOCKER_BUILDKIT=1 docker build -t megaease/easeprobe -f ${MKFILE_DIR}/resources/Dockerfile ${MKFILE_DIR}
--- a/conf/conf_test.go
+++ b/conf/conf_test.go
@@ -28,8 +28,8 @@ import (
 	"testing"
 	"time"
 
-	"bou.ke/monkey"
 	"github.com/megaease/easeprobe/global"
+	"github.com/megaease/easeprobe/monkey"
 	"github.com/megaease/easeprobe/notify"
 	"github.com/megaease/easeprobe/notify/discord"
 	"github.com/megaease/easeprobe/notify/email"
--- a/conf/log_test.go
+++ b/conf/log_test.go
@@ -135,7 +135,8 @@ func TestNonSelfRotateLog(t *testing.T) {
 }
 
 func TestOpenLogFail(t *testing.T) {
-	file := "/dev"
+	// User a directory to cause failure
+	file := os.TempDir()
 
 	l := NewLog()
 	l.File = file
--- a/conf/log_test.go
+++ b/conf/log_test.go
@@ -24,7 +24,7 @@ import (
 	"reflect"
 	"testing"
 
-	"bou.ke/monkey"
+	"github.com/megaease/easeprobe/monkey"
 	log "github.com/sirupsen/logrus"
 	"github.com/stretchr/testify/assert"
 	"gopkg.in/natefinch/lumberjack.v2"
--- a/conf/log_test.go
+++ b/conf/log_test.go
@@ -21,10 +21,8 @@ import (
 	"fmt"
 	"os"
 	"path/filepath"
-	"reflect"
 	"testing"
 
-	"github.com/megaease/easeprobe/monkey"
 	log "github.com/sirupsen/logrus"
 	"github.com/stretchr/testify/assert"
 	"gopkg.in/natefinch/lumberjack.v2"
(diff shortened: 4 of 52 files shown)


samanhappy:
Refined: The SIGBUS error still occurs occasionally, as seen [here](https://github.com/megaease/easeprobe/actions/runs/10717708637/job/29717977463#step:5:208), and I haven't found a solution yet.

proditis (commented):
This is a huge change :muscle: , and monkey patching is still kind of foreign to me. Testing and letting you know.

suchen-sci:
What a great PR. We can finally use `macos-latest`. LOL.

suchen-sci:
> Refined: The SIGBUS error still occurs occasionally, as seen [here](https://github.com/megaease/easeprobe/actions/runs/10717708637/job/29717977463#step:5:208), and I haven't found a solution yet.

Try to fix it, then you will be the new monkey king!

samanhappy:
**Bad news:** It appears that the SIGBUS error is related to the Gomonkey library, as discussed in [this issue](https://github.com/agiledragon/gomonkey/issues/146).

**Good news:** I've discovered a new monkey patching library  [bytedance/mockey](https://github.com/bytedance/mockey). After replacing `gomonkey` with it, the test results are looking promising so far.

proditis (commented):
I have run the tests a few times on my system and i seem to be getting some random errors. I havent pinpointed the cause yet just letting you know in case this something you've seen also. 

See attached image, these are two consecutive `make test` runs with no change between invocations. 
image

samanhappy:
@proditis Thanks so much for your testing efforts! The random errors are also unacceptable to me. To help us investigate further, could you please provide the following information about your test environment:

1. What is your system version? Is it macOS ARM64?
2. What are the exact error messages when the tests fail? Do the errors have anything in common, or are they entirely random?

proditis:
My system is Linux x86_64 and go version go1.23.0 linux/amd64.

I will repeat the tests to try to figure out what is causing this failure. Will keep you posted

proditis:
I took a closer look at the failed test. Do we need to monkey.patch this call? 
image

I think that with a bit of rework on this test case we may be able to get rid of this error.

samanhappy:
Sorry for the delayed response. I reviewed the two error logs and the original `TestOpenLogFail` code. Here are some thoughts:

Both failures occur within the lumberjack library's internal functions. We need to determine whether these errors are caused solely by lumberjack or if our patch operations are contributing to the issue.

While the patch call in the image you attached is necessary, the patch call starting at line 188 (shown below) doesn't seem to have any effect on our test:
```go
var lum *lumberjack.Logger
monkey.PatchInstanceMethod(reflect.TypeOf(lum), "Rotate", func(_ *lumberjack.Logger) error {
    return fmt.Errorf("error")
})
```

I couldn't reproduce the error on my MacBook. My suggestion is to remove this patch code and run the tests again on your system to determine whether the issue originates from the library or the patch. Does that sound like a good plan?

proditis:
Hi thanks for the feedback, although i still dont understand why these calls need to be patched :rofl: 

I commented out the following
```go
	//var lum *lumberjack.Logger
	//	monkey.PatchInstanceMethod(reflect.TypeOf(lum), "Rotate", func(_ *lumberjack.Logger) error {
	//		return fmt.Errorf("error")
	//	})
```

Rerun the tests and got the following error [test3.log](https://github.com/user-attachments/files/16941356/test3.log)

PS: This is how I run my tests in case it helps, i am getting a failed run every ~20-40 runs
```sh
fail=0; runs=1; while [ "$fail" -eq 0 ];do echo "[$runs] Running tests"; make test >test.log || ((fail=fail+1)); ((runs=runs+1)); done
```

proditis (approved):
LGTM :tada: :tada: :tada: :tada:

samanhappy:
Many thanks to @proditis for helping us resolve the random errors. @suchen-sci, you can now review this PR again.

suchen-sci (approved):

suchen-sci:
Sorry for the delayed response... LGTM!

samanhappy:
Oops, the `SIGBUS: bus error` still exists as show [here](https://github.com/megaease/easeprobe/actions/runs/11265981951/job/31328712886#step:5:348) !

proditis:
> Oops, the `SIGBUS: bus error` still exists as show [here](https://github.com/megaease/easeprobe/actions/runs/11265981951/job/31328712886#step:5:348) !

I dont have a mac to run tests, but i suspect the same errors will come up with OpenBSD as well, so i'll prepare a system to run tests and see what fails. OpenBSD is much more strict with these types of violations, so hopefully we'll be able to catch most of these errors.

Could this failure be logger related also?

PS: Maybe we need to open an issue to keep track of these crashes and close it only after we have run enough tests to ensure we fixed all such failures?

samanhappy:
I'm not sure if these errors are related to the logger, but all the `SIGBUS: bus error` occurrences seem to be specific to macOS-arm64.

I've opened an issue as per your advice. Let's track the progress there and keep each other updated on any developments.

PR #642, opened 2025-03-23
Limit redirects

jonhadfield:
Firstly, thanks for creating this application.

The easeprobe http probe has no redirect limit, so testing that a URL is redirecting properly isn't possible. 
For example, I maintain a platform that simply redirects requests for the apex domain to a full url, e.g.
requests for:
https://example.com/shop
should redirect users to the equivalent URL with a 301 status:
https://www.example.com/shop

I've had to include two additional options to make this possible:
```
limit_redirects: <boolean> # needed as max_redirect value of 0 is valid
max_redirects: <integer> # how many response locations to follow before assessing result
```

I think a useful future change would be to assess response headers (Location may be useful for the above).

### DIFF
--- a/docs/Manual.md
+++ b/docs/Manual.md
@@ -337,6 +337,8 @@ http:
     success_code:
       - [200,206] # the code >=200 and <= 206
       - [300,308] # the code >=300 and <= 308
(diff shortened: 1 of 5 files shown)


samanhappy (approved):
LGTM

PR #664, opened 2025-07-16
Add Jitter time to the Probe Interval

qdongxu:
In some special cases. e.g. try to probe a group of endpoints to a single gateway.  It may timeout from time to time as the site detects or considers it's spider or something.

1. jitter the probe interval in the range of (d.ProbeTimeInterval * 0.5, d.ProbeTimeInterval * 1.5).

2. add a Global Option `jitter` (default as false) to keep backward compatibility.


qdongxu:
Close this PR for now as the root cause is not identified undisputedly yet.

